PR #350, opened 2023-01-13
Plug refund bloc

ademar111190:
The progress dialog is not closing as we do not have a refund event. i.e. there is no `BreezEvent` to listen to know when the refund is done; the idea is to have one, or is the refund supposed to have a different flow @erdemyerebasmaz @roeierez ?

### DIFF
--- a/lib/bloc/refund/refund_bloc.dart
+++ b/lib/bloc/refund/refund_bloc.dart
@@ -2,7 +2,10 @@ import 'dart:async';
 
 import 'package:breez_sdk/breez_bridge.dart';
 import 'package:c_breez/bloc/refund/refund_state.dart';
+import 'package:fimber/fimber.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
+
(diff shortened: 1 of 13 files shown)


ademar111190:
> Could you open a separate PR for translations? As for the Refund changes see the TODO items on WaitBroadcastDialog.

https://github.com/breez/c-breez/pull/356

ademar111190:
> > The progress dialog is not closing as we do not have a refund event. i.e. there is no `BreezEvent` to listen to know when the refund is done; the idea is to have one, or is the refund supposed to have a different flow @erdemyerebasmaz @roeierez ?
> 
> @ademar111190 The refund returns the broadcasted tx id so we don't need an event. If it returns the tx is with success then we can close the dialog.

I did update the PR to use the txId returned, I thought that would be a kind of key for an event. Thanks for the information :)

erdemyerebasmaz (approved):
LGTM 🎉 
Let's handle UI changes in another PR as we've discussed.

PR #366, opened 2023-01-18
Fix double minus signal

ademar111190:
|Before|After|
|---|---|
|Screenshot_20230118_095327|Screenshot_20230118_101521|

### DIFF
--- a/lib/routes/home/widgets/payments_list/payment_item_amount.dart
+++ b/lib/routes/home/widgets/payments_list/payment_item_amount.dart
@@ -34,7 +34,7 @@ class PaymentItemAmount extends StatelessWidget {
         builder: (context, currencyState) {
           final fee = _paymentInfo.feeMsat ~/ 1000;
           final amount = currencyState.bitcoinCurrency.format(
-            _paymentInfo.amountMsat ~/ 1000,
+            (_paymentInfo.amountMsat ~/ 1000).abs(),
             includeDisplayName: false,
           );
           final feeFormatted = currencyState.bitcoinCurrency.format(


erdemyerebasmaz (approved):
Great catch! 
Have you checked if we have the same mistake on breezmobile?

lib/routes/home/widgets/payments_list/payment_item_amount.dart
```diff
@@ -34,7 +34,7 @@ class PaymentItemAmount extends StatelessWidget {
         builder: (context, currencyState) {
           final fee = _paymentInfo.feeMsat ~/ 1000;
           final amount = currencyState.bitcoinCurrency.format(
-            _paymentInfo.amountMsat ~/ 1000,
+            (_paymentInfo.amountMsat ~/ 1000).abs(),
```
roeierez, line 37: @ademar111190 so it seems that we decided to use absolute values for PaymentType.Received & PaymentType.Sent.  Perhaps the same should be for PaymenetType.ClosedChannel? Meaning we should fix the amount in the sdk to be absolute?

ademar111190: @roeierez I believe it is a better approach to always use positive numbers because financial transactions always have a negative flow for one side and a positive flow to another side, so how to display or how to calculate should be the responsibility of the code dealing with such numbers.

e.g. which looks better?

```
balance = balance + feePaid;
```
or 
```
balance = balance - feePaid;
```

The second is better to understand when you are reading the code IMHO.

@erdemyerebasmaz It has been a while since I tested this flow on breezmobile, but the last time I tested it (when I did move the texts to the translations files) it was working as expected.

roeierez: @ademar111190 exactly, this is why I thought the `_paymentInfo.amountMsat` should come positive from the SDK. Isn't it aligned with what you wrote above?

ademar111190: Yes it is, I'll close this PR to make this change on the SDK

PR #367, opened 2023-01-18
Fix send to btc title

ademar111190:
Send to BTC flow has 2 possible titles, one when the user spontaneously send its sats to a BTC address and another when he needs to refund after a channel closed, this PR I add a way to correctly set the title on send to BTC flow

### DIFF
--- a/lib/user_app.dart
+++ b/lib/user_app.dart
@@ -194,8 +194,9 @@ class UserApp extends StatelessWidget {
                                   );
                                 case '/withdraw_funds':
                                   return FadeInRoute(
-                                    builder: (_) =>
-                                        const WithdrawFundsAddressPage(),
+                                    builder: (_) => WithdrawFundsAddressPage(
+                                      arguments: settings.arguments as WithdrawFundsArguments,
+                                    ),
                                     settings: settings,
                                   );
                               }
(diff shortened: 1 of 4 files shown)


erdemyerebasmaz:
I suggest we solve this by passing title argument to the named route as such:
https://docs.flutter.dev/cookbook/navigation/navigate-with-arguments

lib/user_app.dart
```diff
@@ -193,10 +193,18 @@ class UserApp extends StatelessWidget {
                                     builder: (_) => QRScan(),
                                     settings: settings,
                                   );
-                                case '/withdraw_funds':
+                                case '/withdraw_funds_spontaneous':
```
roeierez, line 196: good catch! Actually spontaneous sounds a bit misleading here because it sounds like keysend and here we only want to send lightning funds  and swap them with onchain.
Maybe use:
`unexpected_funds `- for the case where we have funds in the onchain wallet
`withdraw_funds` - for the case where we want to send to btc address.

Saying all this I think we need to disable the "Send to btc address" for now as it is not implemented in the sdk yet.

ademar111190: You are right; I did change the names to your suggestion, which sounds better.
I did disable the send to BTC address too.
@erdemyerebasmaz I did apply your suggestion as well

roeierez (approved):
LGTM!

PR #370, opened 2023-01-18
Move translations to breez translations repo

ademar111190:
My idea is to share the same translation files between cbreez and breezmobile to avoid duplicate work and to have up-to-date translations for both projects.

If you agree with this approach, I'll make the breez translations repo public and move the breezmobile translations there as I'm doing here.

Breez Translations repo: https://github.com/breez/Breez-Translations


erdemyerebasmaz (approved):
🙇

PR #371, opened 2023-01-18
Fix iOS build

roeierez:
The fix includes:
* Add entitlement file
* GoogleService-Info.plist file
* Firebase late initialization
* Disable strip symbols on release

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,4 @@ conf/breez.conf
 build-output/
 google-services.json
 ios/Runner/GoogleService-Info.plist
+build-ios.sh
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -48,3 +48,5 @@ app.*.map.json
 conf/breez.conf
 build-output/
 google-services.json
+ios/Runner/GoogleService-Info.plist
+build-ios.sh
```
erdemyerebasmaz, line 52: Is the extension wrong on this file or is it different from build-ios.yml?

roeierez: I have a script for building in my computer so just excluded it.

PR #372, opened 2023-01-18
Disable sweep if user cannot afford fees

erdemyerebasmaz:


erdemyerebasmaz:
@kingonly This is a temporary solution to prevent users from confirming sweep tx's they cannot afford.

https://user-images.githubusercontent.com/4012752/213297401-72891e1c-2575-4558-b781-aa8d0e2ed437.MOV

It's a bit ugly imo but the UI code needs a bit of tidying up to fit a proper solution, which should address these items:
- Fallback to any affordable processing speed, 
  - Display an error message if none can be afforded by the user.
- Disable clicking processing speed if it can't be afforded

kingonly:
@erdemyerebasmaz thanks, it doesn't look bad, but I need to understand what's the solution in breezmobile first. I'm not sure why we need something special for c-breez. 
Btw, the warning message should use the same style we use for fees.

erdemyerebasmaz:
> @erdemyerebasmaz thanks, it doesn't look bad, but I need to understand what's the solution in breezmobile first. I'm not sure why we need something special for c-breez.
> Btw, the warning message should use the same style we use for fees.

There are minor differences between breezmobile & c-breez. On breezmobile, the fee is calculated on breez library, on c-breez we calculate it afterwards after getting recommended fees from mempool.

From a brief glance over the code, we only display Confirm button if API returns FeeOptions. If API returns none, we display a full page "The amount is too small to broadcast. Please try again later" message. If there's at least one, it fills the other remaining FeeOptions as null and disables them.
We can do all this because fees are pre-calculated. We can match the exact behavior but it requires some UI work as the communication between widgets isn't right for it at the moment.

I'll open a new task for it, this solution is a preventative measure for the test tomorrow.

kingonly:
Thanks. From a product standpoint I think the behavior in breezmobile is correct and in general, from a user standpoint, both should behave the same unless we explicitly decided otherwise.

roeierez:
@erdemyerebasmaz is this ready for review? If so please run the CI because currently it was cancelled.

erdemyerebasmaz:
> @erdemyerebasmaz is this ready for review? If so please run the CI because currently it was cancelled.

CI completed with success

roeierez (approved):

Issue #414, opened 2023-01-21
Wrong avatars


kingonly:
If there is a description, we use + or - and not the align avatar. Please review the avatar business logic, including lnurlp, and use the same business logic of Breez mobile.
![Screenshot_2023-01-21-22-40-55-43_73a31df444cba46df764485ac1a8c3ef](https://user-images.githubusercontent.com/31890660/213886349-f61f52e6-ca81-408d-ab71-f0a6a6dda97a.jpg)
![Screenshot_2023-01-21-22-40-40-10_9a3a9060ca72a38f8a1fec1d0c1a640e](https://user-images.githubusercontent.com/31890660/213886350-c5a51dad-0bcc-41cf-b031-fb1608276827.jpg)

PR #432, opened 2023-01-23
Revise avatar icon logic

erdemyerebasmaz:
Addresses https://github.com/breez/c-breez/issues/414

Display +/- icons when payment item 
- has description or
- is sent with keysend or
- is a channel close payment

### DIFF
--- a/lib/routes/home/widgets/payments_list/payment_item_avatar.dart
+++ b/lib/routes/home/widgets/payments_list/payment_item_avatar.dart
@@ -19,7 +19,9 @@ class PaymentItemAvatar extends StatelessWidget {
         radius: radius,
         backgroundColor: Colors.white,
         child: Icon(
-          paymentItem.paymentType == PaymentType.Received ? Icons.add_rounded : Icons.remove_rounded,
+          paymentItem.paymentType == PaymentType.Received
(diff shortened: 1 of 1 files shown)


lib/routes/home/widgets/payments_list/payment_item_avatar.dart
```diff
@@ -29,5 +31,17 @@ class PaymentItemAvatar extends StatelessWidget {
...
   }
 
-  bool get _shouldShowLeadingIcon => paymentItem.details is LnPaymentDetails &&  (paymentItem.details as LnPaymentDetails).keysend;
+  bool get _shouldShowLeadingIcon {
+    final hasDescription = (paymentItem.description != null &&
+        paymentItem.description!.isNotEmpty);
+    final details = paymentItem.details.data;
+    final isKeySend = (details is LnPaymentDetails) ? details.keysend : false;
+    final isClosedChannelPayment =
+        paymentItem.paymentType == PaymentType.ClosedChannel ||
```
roeierez, line 40: PaymentType.ClosedChannel is enough as the condition will return true currently regardless if the channel is pending or closed.

roeierez (changes requested):

PR #433, opened 2023-01-24
change app name to Cloud Breez #390

ubbabeck:
This PR references issue #390. 

The app now displays Cloud Breez when installed on device. 
Cloud Breez

Please let me know if this name change is sufficient or if the scope for the renaming is larger?

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -5,7 +5,7 @@
     <uses-permission android:name="android.permission.NFC"/>  <!-- Deep links -->
 
     <application
-        android:label="c_breez"
+        android:label="Breez Cloud"
         android:name="${applicationName}"
         android:icon="@mipmap/ic_launcher">
         <activity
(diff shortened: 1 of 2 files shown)


kingonly:
Breez Cloud, not Cloud Breez. Did you change the color of the logo as well?

ubbabeck:
Sorry my bad, I fixed it and gave it the correct name now.  I ran the app in debug mode so it was turned into green, here is the proper one.
Breez_Cloud

ubbabeck:
iOS too for reference Screenshot 2023-01-24 at 09 39 53

erdemyerebasmaz:
@ademar111190 Can we add app name parameter to BreezTranslations and replace each app name occurrence with the value of that parameter?

kingonly:
@erdemyerebasmaz but it's a string that apps shouldn't share and by definition we don't want to translate

erdemyerebasmaz:
@kingonly What I had in mind was, for example:
Instead of defining `"status_text_ready"` as `"Breez is ready to receive funds.",`
we define it as `"{appName} is ready to receive funds.",` so we don't need to have a separate translation for Breez Cloud. Which, I'm not sure is possible with .arb files.
A better question would be, do we need to differentiate the app name or is it okay to use 'Breez' for texts on both apps?

kingonly:
Gotcha. We don't need to do it now. It's ok to keep using Breez within Breez Cloud for now.

Issue #412, opened 2023-01-21
Font sizes are smaller in payment details dialog 


kingonly:
Needs to look like Breez Mobile
![Screenshot_2023-01-21-21-11-44-65_9a3a9060ca72a38f8a1fec1d0c1a640e](https://user-images.githubusercontent.com/31890660/213883256-e5f06e69-553f-49f6-bedb-a3be46079718.jpg)
![Screenshot_2023-01-21-21-10-14-59_73a31df444cba46df764485ac1a8c3ef](https://user-images.githubusercontent.com/31890660/213883257-1e64fc3b-6c74-4f3b-a19b-80a5ac82dbb7.jpg)

ademar111190:
I did break the payment dialog widget to find eventual differences https://github.com/breez/c-breez/pull/434 https://github.com/breez/breezmobile/pull/1048

I believe what @kingonly is mentioning as a different font size is, in fact, the "title/subtitle" so I'll study how to have the same behavior with the different models we have on each project.

If you have suggestions @roeierez @erdemyerebasmaz  let me know.

kingonly:
Your might be right. c-breez needs to have the same title subtitle logic of Breez Mobile. There's no reason for a different model.

erdemyerebasmaz:
@ademar111190 can we close this?

ademar111190:
We can close this issue as we have both title and subtitle now. Still, I believe we need to open a new issue to address its content; right now, both fields are duplicated; The only difference from a content perspective is for long texts where the title will trim it, and the subtitle will multiline the content.

erdemyerebasmaz:
@ademar111190 Please go ahead open the issue you mentioned.

Closed.

PR #427, opened 2023-01-23
Update payment title default to Unknown and expand extraction rules

ademar111190:
Related issue: https://github.com/breez/c-breez/issues/401

I did add more ways to extract possible descriptions, and I did set the fallback case to "Unknown"

PR #434, opened 2023-01-24
Break payment dialog widgets

ademar111190:
The idea is to make it easier to find the theming differences reported on https://github.com/breez/c-breez/issues/412


erdemyerebasmaz:
Are there any changes to widgets themselves or were they extracted as is? Asking because it's difficult to review such changes.

About https://github.com/breez/c-breez/issues/412, I don't see any difference in theming on first glance, but we're missing `paymentInfo.dialogTitle` to show a title widget.

ademar111190:
> Are there any changes to widgets themselves or were they extracted as is? Asking because it's difficult to review such changes.
> 
> About #412, I don't see any difference in theming on first glance, but we're missing `paymentInfo.dialogTitle` to show a title widget. Perhaps we should apply the changes you've made [here](https://github.com/breez/c-breez/pull/427) to display a title(on another PR).

No changes, only extracting the widgets to make it easier to compare to breezmobile (I'm extracting the widgets there too)

That is my impression too, so after the extract on breezmobile it will be clear.

erdemyerebasmaz (approved):

Issue #404, opened 2023-01-21
Avatar style of the first item is different 


kingonly:
![Screenshot_2023-01-21-15-53-40-61_9a3a9060ca72a38f8a1fec1d0c1a640e](https://user-images.githubusercontent.com/31890660/213870069-ca946eef-4cdd-40ed-a703-dea359ac4d3d.jpg)

kingonly:
Only after payment. If you open again, it looks the same

PR #435, opened 2023-01-24
Notify PaymentItem when FlipTransition finishes

erdemyerebasmaz:
Fixes https://github.com/breez/c-breez/issues/404

- PaymentItem converted to StatefulWidget
- FlipTransition notifies PaymentItem when it's animation has completed to switch states,
- _createdWithin's return value now reflects it's name, previously it checked if it was not created within the duration

#### Known Issues:
- New item added to PaymentsList doesn't force rebuild,
   - The solution introduced in this PR will be obsolete if this is caused by a state management issue.
- There's a small delay after payment completion waiting for sync operations to finish, if this delay exceeds 10 seconds, the animation won't run.

### DIFF
--- a/lib/routes/home/widgets/payments_list/payment_item.dart
+++ b/lib/routes/home/widgets/payments_list/payment_item.dart
@@ -9,7 +9,7 @@ import 'package:c_breez/routes/home/widgets/payments_list/success_avatar.dart';
 import 'package:c_breez/theme/theme_provider.dart' as theme;
 import 'package:flutter/material.dart';
 
-class PaymentItem extends StatelessWidget {
+class PaymentItem extends StatefulWidget {
   final Payment _paymentInfo;
   final bool _firstItem;
   final GlobalKey firstPaymentItemKey;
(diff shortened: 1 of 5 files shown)


lib/routes/home/widgets/payments_list/payment_item.dart
```diff
@@ -46,29 +60,33 @@ class PaymentItem extends StatelessWidget {
                               blurRadius: 5.0,
                             ),
                           ],
-                        )
-                      : null,
-                  child: _createdWithin(const Duration(seconds: 10))
-                      ? PaymentItemAvatar(_paymentInfo, radius: 16)
-                      : FlipTransition(
+                        ),
+                  child: isPaymentItemNew
```
roeierez, line 64: @erdemyerebasmaz  Does this mean we will get transition for every payment that was created in the last 5 minutes?

erdemyerebasmaz: Yeah, that was pushed by mistake after rebasing. Reduced to 15 seconds.
Added +5 seconds to factor in the delay I mentioned.

Issue #399, opened 2023-01-21
Wrong fee warning message 


kingonly:
Please compare this warning to Breez Mobile. I believe we have different message for an empty wallet.
![Screenshot_2023-01-21-14-11-19-50_9a3a9060ca72a38f8a1fec1d0c1a640e](https://user-images.githubusercontent.com/31890660/213866310-467f05c7-7e2b-420d-a87f-d78a2ad4a5ea.jpg)

kingonly:
Same for Receive via BTC Address. Receive more than 0 sats has no meaning.
![Screenshot_2023-01-21-16-34-30-81_9a3a9060ca72a38f8a1fec1d0c1a640e](https://user-images.githubusercontent.com/31890660/213871661-5623f260-4abd-401b-8deb-8eaf0b18cf89.jpg)

ok300:
I see two ways this issue could be interpreted:

A. It should be 100% identical to the Breez Mobile message, even if this means it will lose extra granularity
* Breez Mobile:
  * _A setup fee of 0.4% will be applied on the received amount._
* c-breez:
  * _A setup fee of 0.4% with a minimum of 2000 sats will be applied for receiving more than 0 sats._
  * 
B. It is fine the way it is, but the issue is fixing the zero amount in empty wallets:
* _... will be applied for receiving more than 0 sats._

@kingonly @roeierez could you please say which one is the one you mean? Thanks.

---

Background info:

In breez-mobile, the template is

> A setup fee of ${lsp.channelFeePermyriad / 100}% will be applied on the received amount.
>
> https://github.com/breez/breezmobile-features/blob/7f16ac1f278fa2f327bd67f9722e9839b6c73097/lib/widgets/lsp_fee.dart#L34

resulting in a message like

> A setup fee of 0.4% will be applied on the received amount.

In c-breez, the template is

> A setup fee of {setUpFee}% with a minimum of {minFee} will be applied for receiving more than {liquidity}.
>
> https://github.com/breez/Breez-Translations/blob/378f1e95d40fb411eca99fd669c9fb1627304bab/lib/app_en.arb#L162

with a resulting message like:

> A setup fee of 0.4% with a minimum of 2000 sats will be applied for receiving more than 0 sats.

kingonly:
There are two templates in Breez Mobile. We also use c-breez one but only when there are existing channels.

This is from Breez Mobile:
[...]

PR #441, opened 2023-01-25
Fix fee warning handling on receive

ok300:
On a fresh setup (no channel open yet), creating an invoice would result in a fee warning like:

> A setup fee of 0.4% with a minimum of 2000 sats will be applied for receiving more than 0 sats.

The zero sat ending was due to the missing channel.

With this PR, the warning is

> A setup fee of 0.4% with a minimum of 2000 sats will be applied on the received amount.

Fixes https://github.com/breez/c-breez/issues/399


ok300:
I also tested the PR with a locally-built apk in both configurations (with and without open channel).

roeierez (approved):

Issue #403, opened 2023-01-21
No flushbar after successful payment


kingonly:

kingonly:
@erdemyerebasmaz let's keep the bugs with the "end user issue". We need to test the use cases after the bugs are fixed.

PR #442, opened 2023-01-25
Fix: show snackbar after payment

ademar111190:
Fixes: https://github.com/breez/c-breez/issues/403


erdemyerebasmaz:
@ademar111190 Hold off with this change before we understand if PaymentResultData, SuccessActionData, accountBloc.paymentResultStream is necessary at all and if we can construct this handler using SDK models & streams.

PR #444, opened 2023-01-26
Use SDK Models for payment result handling

erdemyerebasmaz:
This draft PR is a continuation of #442 and addresses #403

#### Changes
- Differentiate between payment types when displaying payment result flushbar
- Replace SuccessActionData with SuccessAction from SDK
- Rename PaymentResultData -> PaymentResult
- Add payment results of lnurlPay, lnurlWithdraw, sendPayment and sendSpontaneousPayment to paymentResultStream
- Replace wait logic after receiving an event with emitting results with a delay and add logs(by @ademar111190)
- Ran dart format on changed files 
   - We should have the same approach with formatting as we do on breez-sdk and not have any formatting changes in PRs
   
#### SDK Suggestions  @roeierez @ok300 
@ademar111190 and I think it's better to have a paymentResultStream on the SDK as we are already doing with other notifications(event, synced etc) Instead of managing it on AccountBloc. This ensures we'll have the right data at the right time. Before we solve the known issues and work more on this PR we'd like hear your input.

#### Known issues
- Multiple flushbars shown for flows that throw an error in InputHandler
   - One from InputHandler's error handler, one from PaymentResultHandler. We should remove payment result related logic from descendants of InputHandler.

#### Issues encountered
- Flushbar position on iOS looks weird
  - Possible solutions:
    - Display flushbar on top of bottom sheet when on home page, 
    - Deploy a visual trick by using a system navigation bar color that contrasts with bottom sheet(send-qr-receive) background color

### DIFF
--- a/lib/bloc/account/account_bloc.dart
+++ b/lib/bloc/account/account_bloc.dart
@@ -193,7 +193,6 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
       return callbackStatus;
     } catch (e) {
       _log.e("lnurlWithdraw error", ex: e);
-      _paymentResultStreamController.add(PaymentResult(error: e));
       rethrow;
     }
   }
@@ -221,7 +220,6 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
       return lnurlPayResult;
     } catch (e) {
       _log.e("lnurlPay error", ex: e);
-      _paymentResultStreamController.add(PaymentResult(error: e));
       rethrow;
     }
   }
--- a/lib/bloc/account/account_bloc.dart
+++ b/lib/bloc/account/account_bloc.dart
@@ -178,7 +178,7 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
       if (callbackStatus is sdk.LnUrlWithdrawCallbackStatus_Ok) {
         _paymentResultStreamController.add(
           PaymentResult(
-            paymentInfo: state.payments.last,
+            paymentInfo: state.payments.first,
           ),
         );
       }
@@ -204,7 +204,7 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
       );
       _paymentResultStreamController.add(
         PaymentResult(
-          paymentInfo: state.payments.last,
+          paymentInfo: state.payments.first,
           successAction: (lnurlPayResult is sdk.LnUrlPayResult_EndpointSuccess)
               ? lnurlPayResult.data
               : null,
@@ -223,7 +223,7 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
     try {
       await _breezLib.sendPayment(bolt11: bolt11, amountSats: amountSats);
       _paymentResultStreamController.add(
-        PaymentResult(paymentInfo: state.payments.last),
+        PaymentResult(paymentInfo: state.payments.first),
       );
     } catch (e) {
       _log.e("sendPayment error", ex: e);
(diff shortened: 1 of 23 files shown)


roeierez:
> @ademar111190 and I think it's better to have a paymentResultStream on the SDK as we are already doing with other notifications(event, synced etc) Instead of managing it on AccountBloc. This ensures we'll have the right data at the right time. Before we solve the known issues and work more on this PR we'd like hear your input.

I think it makes a lot of sense to add PaymentSent event to the sdk.

lib/bloc/account/account_bloc.dart
```diff
@@ -165,13 +167,22 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
...
     try {
-      return _breezLib.lnurlWithdraw(
+      final callbackStatus = await _breezLib.lnurlWithdraw(
         amountSats: amountSats,
         reqData: reqData,
         description: description,
       );
+      if (callbackStatus is sdk.LnUrlWithdrawCallbackStatus_Ok) {
+        _paymentResultStreamController.add(
+          PaymentResult(
```
roeierez, line 177: @erdemyerebasmaz  We have "InvoicePaid" event for incoming payments and we want to add also "PaymentSent" event as you suggested but it will not fire for incoming payments....

erdemyerebasmaz: We can use LnUrlWithdrawCallbackStatus for lnurl withdraw.

erdemyerebasmaz:
- SDK's payment result events are now listened and passed to app's payment result stream,
- Prior payment result related logic from sendPayment & sendSpontaneousPayment are removed,
- Flushbars for payments that throw PaymentFailed are not handled by InputBloc to prevent showing multiple flushbars same error,
- _extractInnerErrorMessage by default for all exceptions processed through extractExceptionMessage
  - **Reason**: PaymentFailed's error message isn't considered a FFIException, which breaks extractExceptionMessage logic. - @roeierez 
  - **Concerns**: This may break error messages for other use cases.

lib/bloc/account/account_bloc.dart
```diff
@@ -186,11 +204,20 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
...
     _log.v("lnurlPay amount: $amount, comment: '$comment', reqData: $reqData");
     try {
-      return _breezLib.lnurlPay(
+      final lnurlPayResult = await _breezLib.lnurlPay(
         userAmountSat: amount,
         reqData: reqData,
         comment: comment,
       );
+      _paymentResultStreamController.add(
+        PaymentResult(
```
roeierez, line 213: Doesn't it make more sense to handle the success action at the caller of lnurlPay and leave the PaymentResult stream for pure lightning success/failed payments?
This way we won't need to differentiate between these in the UI handler.

roeierez (changes requested):
Looks good, see my comment regarding success action.

lib/bloc/account/account_bloc.dart
erdemyerebasmaz: Yes, I was thinking of making that distinction next.

erdemyerebasmaz: Done. https://github.com/breez/c-breez/pull/444/commits/f6586ea67b0abd18c5e9cda48e91b655bd5086e5 https://github.com/breez/c-breez/pull/444/commits/f74fddd178ac95bd3ab66303df15f63ab1afdd80

roeierez (approved):

PR #445, opened 2023-01-26
Uses the dialog title theme

ademar111190:
I'm using the title style on the dialog theme to address https://github.com/breez/c-breez/issues/412

Breez's `PaymentInfo` model has two fields, `dialog title` and `description`, that we present using two different styles, title and description. C-breez `Payment` model has a description field, but it doesn't have anything else on itself or its attached fields.

I would like to know if you have another solution in mind @roeierez @erdemyerebasmaz; so far, what I have done here is to remove the description style and uses the title style, despite the title content coming from the description field.


lib/routes/home/widgets/payments_list/dialog/payment_details_dialog_amount.dart
```diff
@@ -7,12 +7,12 @@ import 'package:c_breez/models/currency.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 
-class PaymentDetailsAmount extends StatelessWidget {
```
ademar111190, line 10: Just a minor fix on the name to fit breezmobile component and its file name

lib/routes/home/widgets/payments_list/payment_details_dialog.dart
```diff
@@ -36,10 +36,10 @@ Future<void> showPaymentDetailsDialog(
             mainAxisSize: MainAxisSize.min,
             crossAxisAlignment: CrossAxisAlignment.stretch,
             children: [
-              PaymentDetailsDialogDescription(
```
erdemyerebasmaz, line 39: I don't think the description widget should be removed, that was not what I had in mind. I was suggesting we use the extractTitle method you introduced on title widget and display the full description on description widget.

lib/routes/home/widgets/payments_list/dialog/payment_details_dialog_amount.dart
erdemyerebasmaz: I suggest we use shorter and singular names for items we display on PaymentDetailsDialog in the form of: PaymentDetail{Parameter}.
_PaymentDetailTitle, PaymentDetailPreimage, PaymentDetailAmount etc._

ademar111190: I want to keep the same name on BreezMobile. There are more widgets, and the more verbose name has a reason to differentiate from other widgets

erdemyerebasmaz: Last time I checked these were not extracted as standalone widgets on Breezmobile but I see what you mean, it's just a bit tiring on the eyes that's all :) They can stay as they are.

lib/routes/home/widgets/payments_list/dialog/payment_details_dialog_description.dart
```diff
@@ -0,0 +1,44 @@
...
+    super.key,
+    required this.paymentInfo,
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    final texts = context.texts();
+    final themeData = Theme.of(context);
+
+    final description = paymentInfo.extractTitle(texts);
```
erdemyerebasmaz, line 20: I was suggesting to use description as it is here.

lib/routes/home/widgets/payments_list/payment_details_dialog.dart
ademar111190: I just added the subtitle back; it is showing the same text as the title as it is coming from the same property, the only differences are the style and the ellipses/multiline strategy to long texts.

lib/routes/home/widgets/payments_list/dialog/payment_details_dialog_description.dart
ademar111190: Yep, I did it.

lib/routes/home/widgets/payments_list/dialog/payment_details_dialog_amount.dart
ademar111190: You are right; the PR is still open https://github.com/breez/breezmobile/pull/1048

If you want to suggest names there, I can apply the same names here later; the only thing is there are more widgets so the naming can be thought with better context.

erdemyerebasmaz (approved):

PR #447, opened 2023-01-26
Swap in progress

roeierez:
This PR fixed the user experience in 'Receive via BTC address' flow.
We now:
1. Allow only one inprogress swap at a time and at most one "unused" swap address at a time.
2. Aware of mempool transactions using polling the in progress swap.

depends on https://github.com/breez/breez-sdk/actions/runs/4015673422

### DIFF
--- a/lib/routes/subswap/swap/swap_page.dart
+++ b/lib/routes/subswap/swap/swap_page.dart
@@ -1,11 +1,14 @@
 import 'package:breez_sdk/breez_bridge.dart';
 import 'package:breez_sdk/bridge_generated.dart';
 import 'package:breez_translations/breez_translations_locales.dart';
+import 'package:c_breez/bloc/swap_in/swap_in_bloc.dart';
+import 'package:c_breez/bloc/swap_in/swap_in_state.dart';
+import 'package:c_breez/routes/subswap/swap/widgets/inprogress_swap.dart';
 import 'package:c_breez/services/injector.dart';
-import 'package:c_breez/utils/exceptions.dart';
 import 'package:c_breez/widgets/back_button.dart' as back_button;
 import 'package:c_breez/widgets/single_button_bottom_bar.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
 
 import 'widgets/address_widget_placeholder.dart';
 import 'widgets/deposit_widget.dart';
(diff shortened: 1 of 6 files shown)


lib/routes/subswap/swap/widgets/inprogress_swap.dart
```diff
@@ -0,0 +1,69 @@
...
+  const SwapInprogress({
+    required this.swap,
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    final texts = context.texts();    
+    final List<String> swapTransactions = [];
+    swapTransactions.addAll(swap.unconfirmedTxIds);
+    swapTransactions.addAll(swap.unconfirmedTxIds);
```
erdemyerebasmaz, line 20: Why the double call here? Is it by mistake or a workaround for a bug?
Also, since swap isn't being used elsewhere, I suggest we populate the swapTransactions where we define SwapInprogress widget.
`SwapInprogress(swapTransactions: swapState.inProgress!.unconfirmedTxIds.toList())`

!!

I've checked the respective breez-sdk PR and see that unconfirmedTxIds is already a List<String>. We can use it directly.
`SwapInprogress(swapTransactions: swapState.inProgress!.unconfirmedTxIds)`

lib/routes/subswap/swap/widgets/inprogress_swap.dart
```diff
@@ -0,0 +1,69 @@
...
+              ),
+              child: _linkLauncher(context, swapTransactions[0]),
+            ),
+          ],
+        )
+      ],
+    );
+  }
+
+  Widget _linkLauncher(BuildContext context, String unconfirmedTxID) {
```
erdemyerebasmaz, line 54: I don't see the necessity of the helper method here, let's use LinkLauncher in place of it.

lib/user_app.dart
```diff
@@ -160,13 +156,15 @@ class UserApp extends StatelessWidget {
                                   );
                                 case '/swap_page':
                                   return FadeInRoute(
-                                    builder: (_) => const SwapPage(),
+                                    builder: (_) => BlocProvider(
+                                      create: (BuildContext context) => SwapInBloc(ServiceInjector().breezLib),
+                                      child: const SwapPage(),
+                                    ),
```
erdemyerebasmaz, line 162: We add BlocProviders under [MultiBlocProvider](https://github.com/breez/c-breez/blob/ae9bf51a9865df394269c17507d4a7f42360ee11/lib/main.dart#L52) on top level.

lib/routes/subswap/swap/swap_page.dart
```diff
@@ -21,70 +24,56 @@ class SwapPage extends StatefulWidget {
...
-      setState(() {});
-    }
-  }
-
   @override
   Widget build(BuildContext context) {
-    final texts = context.texts();
-    final bool isLoading = swap == null && errorMessage == null;
+    return BlocBuilder<SwapInBloc, SwapInState>(builder: (context, swapState) {
+      final SwapInBloc swapInBloc = context.read();
```
erdemyerebasmaz, line 35: Please define expected type `<SwapInBloc>`

erdemyerebasmaz (changes requested):
This is looking good!
I have left a few comments.

lib/routes/subswap/swap/widgets/inprogress_swap.dart
roeierez: I did a mistake indeed. Should be a concatenation  of both `unconfirmedTxIds` and `confirmedTxIds` since we consider it in progress as long as has utxos (confirmed or unconfirmed).
I will fix.

lib/user_app.dart
roeierez: This is a different kind of bloc that needs to be bound to the page lifecycle (closed when the page is disposed) so the best practice is to wrap the widget that needs it, see here: https://pub.dev/documentation/flutter_bloc/latest/flutter_bloc/Bloc/close.html

lib/routes/subswap/swap/swap_page.dart
roeierez: Why? It is inferred from the type.

lib/routes/subswap/swap/widgets/inprogress_swap.dart
roeierez: @erdemyerebasmaz  I pushed changes which also adapt the page to the changes in https://github.com/breez/breez-sdk/pull/84

lib/routes/subswap/swap/swap_page.dart
erdemyerebasmaz: My bad, didn't notice you defined the type here.
I'll go over the usages of this context extension and pick a single convention to help with search indexing.

lib/routes/subswap/swap/widgets/inprogress_swap.dart
erdemyerebasmaz: > I did a mistake indeed. Should be a concatenation of both `unconfirmedTxIds` and `confirmedTxIds` since we consider it in progress as long as has utxos (confirmed or unconfirmed).
> I will fix.

I'm only seeing unconfirmedTxIds being used, it's not using a concatenation of both.

roeierez: Right. I did that because this was the behavior in breez mobile but actually concatenating them adds to the user experience as the user can see the confirmed transactions in case of redeem error, so I pushed that.
Thanks.

erdemyerebasmaz (approved):
LGTM!
Please rerun CI after https://github.com/breez/breez-sdk/pull/81 is merged on breez-sdk before merging this PR.

Issue #387, opened 2023-01-21
No default avatar 


kingonly:
![Screenshot_2023-01-21-11-27-35-70_9a3a9060ca72a38f8a1fec1d0c1a640e](https://user-images.githubusercontent.com/31890660/213860853-7a36597e-70cd-4967-bcd3-046c3ee07e9d.jpg)

PR #449, opened 2023-01-26
Add a profile bloc logic to fill missing properties on user profile

ademar111190:
Fixes https://github.com/breez/c-breez/issues/387

### DIFF
--- a/lib/bloc/user_profile/user_profile_bloc.dart
+++ b/lib/bloc/user_profile/user_profile_bloc.dart
@@ -23,45 +23,51 @@ class UserProfileBloc extends Cubit<UserProfileState> with HydratedMixin {
     this._breezServer,
     this._notifications,
   ) : super(UserProfileState.initial()) {
-    stream.listen((profile) {
-      _log.v("State: ${profile.profileSettings.toJson()}");
-      final settings = profile.profileSettings;
-      if (settings.color == null || settings.animal == null) {
-        _log.v("Profile has not color or name, generating new random ones…");
-        final defaultProfile = generateDefaultProfile();
-        emit(profile.copyWith(
-          profileSettings: settings.copyWith(
-            color: settings.color ?? defaultProfile.color,
-            animal: settings.animal ?? defaultProfile.animal,
-          ),
-        ));
-      } else if (settings.name == null) {
-        _log.v("Profile has no name, generating new one…");
-        emit(profile.copyWith(
-          profileSettings: settings.copyWith(
-            name: DefaultProfile(
-              settings.color!,
-              settings.animal!,
-            ).buildName(getSystemLocale()),
-          ),
-        ));
-      } else if (!settings.registered) {
-        _log.v("Profile has not been registered, registering…");
-        registerForNotifications();
-      }
-    });
-    emit(state); // trigger hydration
+    var profile = state;
+    _log.v("State: ${profile.profileSettings.toJson()}");
+    final settings = profile.profileSettings;
+    if (settings.color == null || settings.animal == null) {
+      _log.v("Profile has not color or name, generating new random ones…");
+      final defaultProfile = generateDefaultProfile();
+      profile = profile.copyWith(
+        profileSettings: settings.copyWith(
+          color: settings.color ?? defaultProfile.color,
+          animal: settings.animal ?? defaultProfile.animal,
+        ),
+      );
+    }
+    if (settings.name == null) {
+      _log.v("Profile has no name, generating new one…");
+      profile = profile.copyWith(
+        profileSettings: settings.copyWith(
+          name: DefaultProfile(
+            settings.color!,
+            settings.animal!,
+          ).buildName(getSystemLocale()),
+        ),
+      );
+    }
+    emit(profile);
+    registerForNotifications();
   }
 
   Future registerForNotifications() async {
     _log.v("registerForNotifications");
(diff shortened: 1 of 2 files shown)


lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -1,33 +1,74 @@
...
   final BreezServer _breezServer;
   final Notifications _notifications;
 
-  UserProfileBloc(this._breezServer, this._notifications)
-      : super(UserProfileState.initial());
+  UserProfileBloc(
+    this._breezServer,
+    this._notifications,
+  ) : super(UserProfileState.initial()) {
+    stream.listen((profile) {
```
roeierez, line 26: Why do we need to listen to the stream? Isn't it enough to look for the current value? e.g `this.state`

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -1,33 +1,74 @@
...
+      if (settings.color == null || settings.animal == null) {
+        _log.v("Profile has not color or name, generating new random ones…");
+        final defaultProfile = generateDefaultProfile();
+        emit(profile.copyWith(
+          profileSettings: settings.copyWith(
+            color: settings.color ?? defaultProfile.color,
+            animal: settings.animal ?? defaultProfile.animal,
+          ),
+        ));
+      } else if (settings.name == null) {
```
roeierez, line 38: why `else if` on first install we won't have color, animal, name and notification registered so we need to fill/generate any of this properties unconditionally.

roeierez (changes requested):

erdemyerebasmaz (commented):
Why don't we just generate a random image and name on UserProfileSettings.initial()'s constructor? This stream logic seems excessive.

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -1,33 +1,74 @@
...
+          profileSettings: settings.copyWith(
+            name: DefaultProfile(
+              settings.color!,
+              settings.animal!,
+            ).buildName(getSystemLocale()),
+          ),
+        ));
+      } else if (!settings.registered) {
+        _log.v("Profile has not been registered, registering…");
+        registerForNotifications();
```
roeierez, line 50: @ademar111190  Also don't we need to register for push notifications on every restart? The push token may change over time.

lib/bloc/user_profile/user_profile_bloc.dart
ademar111190: That would work to fix the profile state at the creation time of the bloc; I did this through the stream to have this fix applied to any state change; if you think it is not necessary I can remove the stream.

lib/bloc/user_profile/user_profile_bloc.dart
ademar111190: Because it is in the stream, it will be filled in three steps; I broke it into small steps to clarify what is being changed and why; maybe it is too much from my side.

ademar111190:
> Why don't we just generate a random image and name on UserProfileSettings.initial()'s constructor? This stream logic seems excessive.

That was my first approach, but we end up with a different profile every run and we didn't fix profiles that already exist when someone updates the app.

Also, I think it is advantageous to have a fix applied because in the future if we add some property to the profile we can "migrate" this model in this very way.

lib/bloc/user_profile/user_profile_bloc.dart
ademar111190: Good point, you are rigth, I'll change the logic here.

lib/bloc/user_profile/user_profile_bloc.dart
ademar111190: Stream removed

lib/bloc/user_profile/user_profile_bloc.dart
ademar111190: Checking all at one single step (kind of, still break each part)

lib/bloc/user_profile/user_profile_bloc.dart
ademar111190: Fixed

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -1,33 +1,80 @@
...
-  UserProfileBloc(this._breezServer, this._notifications)
-      : super(UserProfileState.initial());
+  UserProfileBloc(
+    this._breezServer,
+    this._notifications,
+  ) : super(UserProfileState.initial()) {
+    var profile = state;
+    _log.v("State: ${profile.profileSettings.toJson()}");
+    final settings = profile.profileSettings;
+    if (settings.color == null || settings.animal == null) {
```
roeierez, line 29: It looks like if the color and animal are null then name is null as well so maybe we can generate the name inside this block as well removing the bellow block?

roeierez (approved):
LGTM with a small comment worth checking.

lib/bloc/user_profile/user_profile_bloc.dart
ademar111190: I'm a little bit paranoic, but I want to avoid overriding anything that exists by doing it on the same block I need to use the actual.color ?? new.color and create a new generator, having in separated blocks, I can guarantee no property override at all.

Issue #394, opened 2023-01-21
Multi LSP selection text isn't shown in light mode


kingonly:
"Click here to select LSP" is probably in white

PR #452, opened 2023-01-27
Fix no lsp widget theme

ademar111190:
Fixes https://github.com/breez/c-breez/issues/394

### DIFF
--- a/lib/routes/home/widgets/no_lsp_widget.dart
+++ b/lib/routes/home/widgets/no_lsp_widget.dart
@@ -10,19 +10,22 @@ class NoLSPWidget extends StatelessWidget {
   Widget build(BuildContext context) {
     final texts = context.texts();
     final themeData = Theme.of(context);
+    final foregroundColor = themeData.textTheme.labelLarge!.color!;
 
     return Column(
       crossAxisAlignment: CrossAxisAlignment.center,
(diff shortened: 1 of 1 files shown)


lib/routes/home/widgets/no_lsp_widget.dart
```diff
@@ -10,19 +10,22 @@ class NoLSPWidget extends StatelessWidget {
...
-        Row(
-          mainAxisAlignment: MainAxisAlignment.center,
-          mainAxisSize: MainAxisSize.max,
-          children: [
-            Text(
-              texts.no_lsp_widget_message,
-            )
-          ],
+        Padding(
+          padding: const EdgeInsets.fromLTRB(16, 0, 16, 0),
```
erdemyerebasmaz, line 20:
```suggestion
          padding: const EdgeInsets.symmetric(horizontal: 16, vertical: 0),
```

erdemyerebasmaz (approved):
Approved with a small suggestion.

lib/routes/home/widgets/no_lsp_widget.dart
ademar111190: TBH I prefer fromLTRB, over the only/symmetric/etc…

PR #455, opened 2023-01-30
no clipboard handling for appView and webView handling

ubbabeck:
Remove case webview and appview in order to only react on lighting related stuff and not regular links in clipboard.

### DIFF
--- a/lib/bloc/input/input_bloc.dart
+++ b/lib/bloc/input/input_bloc.dart
@@ -60,8 +60,6 @@ class InputBloc extends Cubit<InputState> {
                 protocol: command.protocol,
                 inputData: command.decoded as LNURLParseResult);
           case breez_sdk.InputProtocol.nodeID:
-          case breez_sdk.InputProtocol.appLink:
-          case breez_sdk.InputProtocol.webView:
             return InputState(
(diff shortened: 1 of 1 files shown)


lib/bloc/input/input_bloc.dart
```diff
@@ -91,7 +89,8 @@ class InputBloc extends Cubit<InputState> {
     return InputState(protocol: command.protocol, inputData: invoice);
   }
 
-  Stream<DecodedClipboardData> get decodedClipboardStream => _device.clipboardStream.distinct().skip(1).map((clipboardData) {
+  Stream<DecodedClipboardData> get decodedClipboardStream =>
+      _device.clipboardStream.distinct().skip(1).map((clipboardData) {
```
erdemyerebasmaz, line 93: Is this a dart format change?

erdemyerebasmaz (approved):
Thanks! You can merge after addressing my comment.

lib/bloc/input/input_bloc.dart
ubbabeck: yes this must have been automatic styling , I only removed the two cases. Do we use some specific style rules?

PR #458, opened 2023-01-31
generate name inside bloc fix null issue

ubbabeck:
Related to https://github.com/breez/c-breez/issues/387.

Build fails due to color and animal being null when trying to set name.

fix by moving logic inside block where color and animal is set ref: https://github.com/breez/c-breez/pull/449#discussion_r1090037121. Do not think this should cause any issues or side effects.

### DIFF
--- a/lib/bloc/user_profile/user_profile_bloc.dart
+++ b/lib/bloc/user_profile/user_profile_bloc.dart
@@ -26,23 +26,18 @@ class UserProfileBloc extends Cubit<UserProfileState> with HydratedMixin {
     var profile = state;
     _log.v("State: ${profile.profileSettings.toJson()}");
     final settings = profile.profileSettings;
-    if (settings.color == null || settings.animal == null) {
+    if (settings.color == null && settings.animal == null) {
       _log.v("Profile has not color or name, generating new random ones…");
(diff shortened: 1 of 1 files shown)


lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -26,23 +26,18 @@ class UserProfileBloc extends Cubit<UserProfileState> with HydratedMixin {
     var profile = state;
     _log.v("State: ${profile.profileSettings.toJson()}");
     final settings = profile.profileSettings;
-    if (settings.color == null || settings.animal == null) {
+    if (settings.color == null && settings.animal == null) {
```
ademar111190, line 29: It needs to be an `or` not an `and`. If you are adding the name to the same if statement you need to add the `name == null` as a third or clause.

ademar111190 (changes requested):

lib/bloc/user_profile/user_profile_bloc.dart
ubbabeck: Added correct checking as well as generating the name properly.

ademar111190 (approved):

PR #464, opened 2023-02-06
Update "Expert Keys" output for sdk-cli compatibility

### LINKED ISSUE #317: Add sql storage to the export keys output
It would provide a good way to debug issues.

ok300:
Fixes #317


erdemyerebasmaz (approved):

lib/bloc/account/account_bloc.dart
```diff
@@ -74,9 +74,10 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
 
   Future _startRegisteredNode() async {    
     final credentials = await _credentialsManager.restoreCredentials();
+    final seed = bip39.mnemonicToSeed(credentials.mnemonic);
```
roeierez, line 77: We need to make sure that we all create new wallets since this is not backward compatible. It is reasonable because we are pre alpha.

ok300: Correct.

I tested it with new wallets, as well as restore previous wallet.

Just upgrading c-breez "in place" will not work. It will probably need the previous app to be uninstalled first.

roeierez (approved):
Everyone that install with this PR needs to restore the wallet or create a new one.

Issue #461, opened 2023-01-31
Payment dialog title and subtitle content


ademar111190:
Context: https://github.com/breez/c-breez/issues/412

Title and subtitle fields are duplicated from a content perspective; The only difference is for long texts where the title will trim it, and the subtitle will multiline the content.

kingonly:
Please replicate the breezmobile logic. There is some duplication there in some cases as well (we have only a payment description), but not in others (when we have the payer/payee information and the description for example).

PR #466, opened 2023-02-08
Replicate description logic from breez mobile on items where data is available

ademar111190:
Fix https://github.com/breez/c-breez/issues/461

Some data is not available on c breez, so we are not 100% equal; reference: https://github.com/breez/breezmobile/blob/232f1178d86b1b7152a03036ab33aafcf0c933a9/lib/bloc/account/account_model.dart#L774


erdemyerebasmaz (approved):
🚀

roeierez (approved):

PR #467, opened 2023-02-10
fix issue with lnurl parcing on close #333

### LINKED ISSUE #333: Support deep links for various payment types
We need to add deep link as we have in Breez wallet for lightning invoices, lnurl protocol, etc...

ubbabeck:
This fixes a race condition where the app did not parse lighting invoice when app where closed (not running in background).

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -31,7 +31,8 @@
             </intent-filter>
 
             <!-- Deep links -->
-            <intent-filter>
+            <meta-data android:name="flutter_deeplinking_enabled" android:value="true" />
+            <intent-filter android:autoVerify="true">
                 <action android:name="android.intent.action.VIEW"/>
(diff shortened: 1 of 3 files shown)


android/app/src/main/AndroidManifest.xml
```diff
@@ -31,7 +31,8 @@
             </intent-filter>
 
             <!-- Deep links -->
-            <intent-filter>
+            <meta-data android:name="flutter_deeplinking_enabled" android:value="true" />
```
erdemyerebasmaz, line 34: Should we add the iOS counterpart of this change on Info.plist? Did not having this broke behavior on Android?

erdemyerebasmaz (approved):
Looks good, thanks for your discovery.
One small comment.

android/app/src/main/AndroidManifest.xml
ubbabeck: Yes agree, I have added it now.

PR #475, opened 2023-02-12
Wait for payment result flushbar to disappear before showing VerifyMnemonicsDialog

erdemyerebasmaz:
Flushbar's default display duration is 8 seconds.


lib/handlers/verify_mnemonics_handler.dart
```diff
@@ -20,20 +20,25 @@ class MnemonicsVerificationHandler {
       streamSubscription = accountBloc.stream
           .where((state) => state.payments.isNotEmpty)
           .distinct((previous, next) => previous.payments.length == next.payments.length)
+          .skip(1)
```
ubbabeck, line 23: Are we skipping the first one because of adding the previous sessions clipboard to persistent storage as explained [here](https://github.com/breez/breezmobile/pull/1050/files#diff-084cdae0036db656c72c770af8a90b016f8fbea85975eb06322924a2f0d58c0dR36) ?

erdemyerebasmaz: No, I should have added comments.

AccountState doesn't get assembled synchronously, this causes a false positive state to pass through this where/distinct filter on launch. That is what we're skipping here.

roeierez: @erdemyerebasmaz is this streams should emit on every time the payment list changes?
Isn't it better (and more explicit) to directly listen to sdk events here? (InvoicePaid, PaymentSucceed)

roeierez: @erdemyerebasmaz yes:
InvoicePaid - for incoming payment.
PaymentSucceed = for outgoing payment.

erdemyerebasmaz:
Closing as requirements has changed. Will open a new PR.

PR #482, opened 2023-02-13
Simplify config files

roeierez:
Took the opportunity to simplify the breez.conf file. Since the specific firebase files are required we don't need these additional parameters in the configuration anymore.
@ubbabeck  I found an in issue with the `_initializeInputBloc` function so took the opportunity to fix that.
Also added documentation for the firebase requirements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,17 @@ c-breez/
 
 ```
 
+### Add firebase config files
+c-breez depends on google services and requires a configured firebase app.
+
+To create your firebase app follow the following link
+[create-firebase-project](https://firebase.google.com/docs/android/setup#create-firebase-project).
+
+After creating the app follow the instructions to create the specific 
+configuration file for your platform:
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -41,6 +41,17 @@ c-breez/
...
+c-breez depends on google services and requires a configured firebase app.
+
+To create your firebase app follow the following link
+[create-firebase-project](https://firebase.google.com/docs/android/setup#create-firebase-project).
+
+After creating the app follow the instructions to create the specific 
+configuration file for your platform:
+* For android - place the google-services.json in the andrpod/app folder
+* For iOS - place the GoogleService-info.plist under ios/Runner folder
+
```
ubbabeck, line 54: Should we add that people can contact us regarding obtaining the API key?

ubbabeck (approved):
Thank you, looks good to me.

PR #485, opened 2023-02-14
Update dependencies

erdemyerebasmaz:
Deprecations
- **flutter_svg**: The color and colorBlendMode properties have been removed. Instead, use the colorFilter property.
- **mobile_scanner**:
  - allowDuplicates is removed and replaced by MobileScannerSpeed enum
  - onDetect now returns the object BarcodeCapture, which contains a List of barcodes and, if enabled, an image.

Important fixes:
- **connectivity_plus**: Do not return ConnectivityResult.none on iOS and MacOS with VPN (#1335).


lib/routes/home/widgets/app_bar/home_app_bar.dart
```diff
@@ -24,7 +24,10 @@ class HomeAppBar extends AppBar {
               "src/icon/hamburger.svg",
               height: 24.0,
               width: 24.0,
-              color: themeData.appBarTheme.actionsIconTheme?.color,
+              colorFilter: ColorFilter.mode(
+                themeData.appBarTheme.actionsIconTheme!.color!,
```
ubbabeck, line 28: Are there any chance that these can be null ?

ubbabeck (approved):
Tested on receiving and scanning on iOS. Sending works and receiving flow is smooth. Does not freeze anymore after receiving funds either.  Approve after addressing the null safety question.

lib/routes/home/widgets/app_bar/home_app_bar.dart
erdemyerebasmaz: Not in our case, they are defined on every theme available on our app.

ubbabeck: Great, thank you 🙏

PR #501, opened 2023-02-21
Fix network message

ademar111190:
I was trying to create a new wallet to debug out the newest bugs, and I got this error, so I'm fixing it before moving to the bug fixes.

I'm keeping the logs I added that helped me to understand the issue; let me know if you prefer me to remove them.

Also, I'm copying the replace error message for the localized ones we did add to breez mobile.

To change the error message, update this entry https://github.com/breez/Breez-Translations/blob/main/lib/app_en.arb#L2278 or send me the better text for me to update it.

|Before|After|
|---|---|
|before|after|

### DIFF
--- a/lib/routes/create_invoice/create_invoice_page.dart
+++ b/lib/routes/create_invoice/create_invoice_page.dart
@@ -20,9 +20,12 @@ import 'package:c_breez/widgets/keyboard_done_action.dart';
 import 'package:c_breez/widgets/receivable_btc_box.dart';
 import 'package:c_breez/widgets/single_button_bottom_bar.dart';
 import 'package:c_breez/widgets/transparent_page_route.dart';
+import 'package:fimber/fimber.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
+
+final _log = FimberLog("CreateInvoicePage");
 
 class CreateInvoicePage extends StatefulWidget {
   final Function(LNURLWithdrawPageResult? result)? onFinish;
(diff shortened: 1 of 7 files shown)


erdemyerebasmaz:
Changes look good, can merge after reviewing the text with @kingonly

kingonly:
Change the first sentence to: "Failed to connect to Breez Cloud."

erdemyerebasmaz:
@ademar111190
In my ideal world we define an appName to l10n delegate once on startup
`BreezTranslations(appName:"BreezCloud")`
and can use it as a global constant on messages
`"generic_network_error" : "Failed to connect to {appName}"`
This ofc, is not possible with intl package & due to limitation of .arb files.

So, I have two suggestions:
1. Multiply each message that has the app name Breez in it's value and add a cloud or cbreez suffix to it's key:
   - `cbreez_generic_network_error` in this case.
2. Have c-breez point to another branch with it's own values but this would make it harder for contributors if Breez Cloud diverges from Breez mobile in the future and also keeping l10n for both apps in sync may prove difficult.

or we move to another internationalization solution with those capabilities, if there's any, I mentioned earlier :)

ademar111190:
@erdemyerebasmaz I believe it is better to have repeated entries; two texts can happen to work in English with a variable like that, but there is no guarantee it will work in other languages.

An interesting example for sharing here is regarding the name of the product "Breez" In Portuguese and Spanish, two languages I did translate the app, we need to know the gender of the subject to write a complete sentence correctly, so I had to ask Roy if Breez is a boy or a girl 🤣 

Latin languages and English are cousins, and even being so close, some things can completely change the phrase, verb tenses, plurals, nouns, pronouns, etc. So imagine when adding translation for Japanese, Chinese, Arabic, and other languages with no similarities to English…

So I think less variable is better, even if that means repeated entries such as that case where we are changing just the name, which I believe is what you propose #1

kingonly:
@erdemyerebasmaz @ademar111190 keep things simple for now. We don't have to use Cloud in these messages. Breez is enough for now.

Issue #499, opened 2023-02-21
Preferences state (collapse) should be persistent 


kingonly:
Close it and restart the app.

![Screenshot_2023-02-21-07-09-28-46_9a3a9060ca72a38f8a1fec1d0c1a640e](https://user-images.githubusercontent.com/31890660/220253163-36d729b6-7df5-4344-8f97-7fef7aaaa241.jpg)

PR #503, opened 2023-02-21
Persist AppMode* and Preferences expansion state

erdemyerebasmaz:
This issue addresses #499 

- Preferences tiles expansion state is now being persisted
- AppMode isn't being used at the moment but it was missing from updateProfile & UserProfileSettings.copyWith.

_This whole drawer logic needs another round of refactoring & simplification imo._

### DIFF
--- a/lib/bloc/user_profile/user_profile_bloc.dart
+++ b/lib/bloc/user_profile/user_profile_bloc.dart
@@ -84,16 +84,22 @@ class UserProfileBloc extends Cubit<UserProfileState> with HydratedMixin {
     String? image,
     bool? registrationRequested,
     bool? hideBalance,
+    AppMode? appMode,
+    bool? expandPreferences,
   }) {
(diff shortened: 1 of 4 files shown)


roeierez:
Do we restore the app mode? I don't see it in the code (you also wrote it is not used).

erdemyerebasmaz:
> Do we restore the app mode? I don't see it in the code (you also wrote it is not used).

We don't restore app mode on breezmobile. 
We do persist the preference between sessions but currently there's only one app mode on c-breez, so it's doesn't really have any purpose.

erdemyerebasmaz:
> Code: Looks good to me.
> 
> Test: Opened and reopened the app. No issues on android emulator.
> 
> Believe this issue must have been android specific since I where not able to recreate it on iOS. Should I start I testing all PRs that affect the UI on both iOS and Android or is it safe to assume that flutter is consistent across OSs?

I don't think it's needed unless specified or the changed widgets belong to a package, where they may be differences.

Issue #409, opened 2023-01-21
Wrong UI for lnurlp


kingonly:
The UI should look like Breez Mobile: title, image location, button, no warning 

![Screenshot_2023-01-21-16-18-21-09_9a3a9060ca72a38f8a1fec1d0c1a640e](https://user-images.githubusercontent.com/31890660/213870993-7ea68dfe-26aa-4ade-a623-8d8345446eca.jpg)
![Screenshot_2023-01-21-16-17-52-54_73a31df444cba46df764485ac1a8c3ef](https://user-images.githubusercontent.com/31890660/213870994-7d518418-df21-4592-a138-41431b4bd9a6.jpg)

erdemyerebasmaz:
Re-assigned to @ubbabeck

kingonly:
The flow is still different from Breez mobile.
After I enter the amount, the user is on the lnurlp screen instead of getting a dialog and showing the processing payment on the home screen like in Breez Mobile.

PR #504, opened 2023-02-22
Show processing payment dialog while lnurl payment is in place

erdemyerebasmaz:
This PR addresses https://github.com/breez/c-breez/issues/409#issuecomment-1437853674

LNURL payment related pages no longer handle payment & pay result handling and instead pop with payment info.
Then a Processing Payment dialog is shown that uses this payment info and handle payment.
Once payment is handled, this dialog is popped with LnUrlPayResult, which is handled by handlePayRequest as usual.

## Test
- [x] LnUrlPayResult_EndpointSuccess
- [x] LnUrlPayResult_EndpointError
- [x] Unknown response from lnurlPay
- [x] Error sending LNURL payment


erdemyerebasmaz:
The payment function error isn't being handled properly at the moment and _handlingRequest doesn't reset.

erdemyerebasmaz:
> The payment function's error isn't being handled properly at the moment and _handlingRequest doesn't reset, preventing handling other payments.

Resolved error handling https://github.com/breez/c-breez/pull/504/commits/f1a19ce609ae48cfe81e8d7ab403c00612931ae5

Also noticed amount wasn't being displayed in any shape or form for fixed amount widgets that allowed comments. Addressed by displaying amount in a readOnly-disabled state https://github.com/breez/c-breez/pull/504/commits/ddec0c9781c9aac460397edf1cd58212dc083870

lib/routes/lnurl/lnurl_invoice_delegate.dart
```diff
@@ -72,14 +81,47 @@ Future<LNURLPaymentPageResult?> handlePayRequest(
...
+        return LNURLPaymentPageResult(
+          successAction: result.data,
+        );
+      } else if (result is sdk.LnUrlPayResult_EndpointError) {
+        _log.v("LNURL payment failed: ${result.data.reason}");
+        return LNURLPaymentPageResult(
+          error: result.data.reason,
+        );
+      }
+    } else if (result is FfiException) {
```
roeierez, line 116: Perhaps it is better to check for any exception here, not only ffi?

lib/routes/lnurl/payment/processing_lnurlp_dialog.dart
```diff
@@ -0,0 +1,145 @@
+import 'dart:async';
+
+import 'package:c_breez/widgets/payment_dialogs/processing_payment/processing_payment_animated_content.dart';
+import 'package:c_breez/widgets/payment_dialogs/processing_payment/processing_payment_content.dart';
+import 'package:flutter/material.dart';
+
+const _kPaymentListItemHeight = 72.0;
+
+class ProcessingLNURLPaymentDialog extends StatefulWidget {
```
roeierez, line 9: This seems like a duplication of the processing payment dialog we have right now. What is the advantage of this rather than use the existing one?

roeierez (changes requested):

lib/routes/lnurl/lnurl_invoice_delegate.dart
erdemyerebasmaz: Our extractExceptionMessage logic are suited for FfiException's and looking at the previous logic I assumed we returned a generic error for other unknown errors.

lib/routes/lnurl/payment/processing_lnurlp_dialog.dart
erdemyerebasmaz: I believe ProcessingPaymentDialog will get complicated and hard to debug if we try to inject this logic there.

ubbabeck (commented):
Tested and the flow and the dialog shows, but we need to being able to approve/ cancel the payment. like in breezmobile. But this might me in the workings 😄 

Code: Looks good to me.

erdemyerebasmaz:
> Tested and the flow and the dialog shows, but we need to being able to approve/ cancel the payment. like in breezmobile. But this might me in the workings 😄
> 
> Code: Looks good to me.

@roeierez Can we cancel LNURL payments?

lib/routes/lnurl/payment/processing_lnurlp_dialog.dart
erdemyerebasmaz: Merged here https://github.com/breez/c-breez/pull/504/commits/88ef255e3f98531133e208d41911ed1b38a16a8f

lib/routes/lnurl/lnurl_invoice_delegate.dart
roeierez: Yes but the problem it is now depends on ffi here at the high level code which doesn't seem clean.
Maybe it is better to make it generic?
For example we can add optional default error parameter to the  `extractExceptionMessage` in case it fails to find what it is looking for (such as ffiexception).
Something like:
`extractExceptionMessage(Object exception, String defaultErrorMsg)`
Or we can find other pattern as long as we don't handle this detail in high level.

erdemyerebasmaz: Done 90ba55faf48ab9125368595a6f98a50e1452aa26

roeierez:
> @roeierez Can we cancel LNURL payments?

You mean before the lightning payment started? Yes otherwise what is the point in the dialog. Let's do the same as breez mobile.

erdemyerebasmaz:
> > @roeierez Can we cancel LNURL payments?
> 
> You mean before the lightning payment started? Yes otherwise what is the point in the dialog. Let's do the same as breez mobile.

Afterwards.

ubbabeck:
> > > @roeierez Can we cancel LNURL payments?
> > 
> > 
> > You mean before the lightning payment started? Yes otherwise what is the point in the dialog. Let's do the same as breez mobile.
> 
> Afterwards.

Shouldn't we not attempt to send the payment before the user has pressed the approve dialog in the popup and all that the lnurp page is redirect the relevant information and then prompt the user with **cancel** or **approve** or have I missed something?

PR #511, opened 2023-02-24
Add domain to request data

ademar111190:
The domain field was added to breez SDK, but it is missing on c-breez; I'm adding it to fix the build.


ok300 (approved):
Was not aware of that section, thanks!

ademar111190:
> Was not aware of that section, thanks!

we are here to help each other 😘

Issue #506, opened 2023-02-22
Extract error message


erdemyerebasmaz:
> LNPay.co error: LNPay.co error 400: Unable to pay LN Invoice: FAILURE_REASON_NO_ROUTE

This error should be extracted and beautified if necessary.

Perhaps use "Unable to pay LN Invoice:" + paymentErrorsMapping logic from breez mobile for the remainder part.

PR #512, opened 2023-02-24
Extract Ln payments error messages

ademar111190:
Fixes: https://github.com/breez/c-breez/issues/506

Pulls the payment error messages from breez mobile.

Waiting for https://github.com/breez/c-breez/pull/504 merges to make a real-world test; if the message is a little bit different from the messages I did write on the unit test which I take from the issue, a minor change will be needed on the extraction logic.

### DIFF
--- a/lib/bloc/account/payment_result.dart
+++ b/lib/bloc/account/payment_result.dart
@@ -14,7 +14,7 @@ class PaymentResult {
 
   String errorMessage(CurrencyState currencyState) {
     final texts = getSystemAppLocalizations();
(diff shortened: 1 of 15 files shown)


roeierez (approved):

erdemyerebasmaz:
@ademar111190 Please re-run CI before merging. The previous error about domain was resolved but there were formatting issues found on the next steps.
We now use 110 line width so either use `dart format --fix -l 110 .` before committing or 
enable lefthook via `lefthook install`. More details [here](https://github.com/breez/c-breez#pre-commit-dart-format-with-lefthook).
Thank you 🙏

Issue #489, opened 2023-02-20
QR button is disabled after fresh install 


kingonly:
![Screenshot_2023-02-20-10-04-30-35_9a3a9060ca72a38f8a1fec1d0c1a640e](https://user-images.githubusercontent.com/31890660/220047721-4914e6c6-3283-46ae-b2d7-f591a0dea2c4.jpg)

Click or focus change fixes it

roeierez:
@erdemyerebasmaz I suspect it might be an invisible dialog/mast that is shown (maybe backup prompt)?

erdemyerebasmaz:
@roeierez No, there isn't actually a bug here. @kingonly QR button and send/receive options are disabled until account is connected to a LSP. If you click Send/Receive on startup, their options should be disabled as well.

Maybe we should grey out send & receive buttons or their text until then?

PR #514, opened 2023-02-27
ensure selected LSP, defer checking connection to lsp

ubbabeck:
# About pr

Since it can take some time to establish a connection the the LSP we first checking if the user has selected a LSP. 

This should solve the issue where the QR button looks as if it disabled ref #489. 

I do not have to much experience with state management in flutter and are not sure if I have followed best practices and would appreciate if you could review this too.

We most likely can defer checking if the _swapper_ **Receive via BTC** is connected to the LSP until a later state of the app. Not sure where to do this.

### DIFF
--- a/lib/routes/create_invoice/create_invoice_page.dart
+++ b/lib/routes/create_invoice/create_invoice_page.dart
@@ -7,7 +7,6 @@ import 'package:c_breez/bloc/currency/currency_bloc.dart';
 import 'package:c_breez/bloc/currency/currency_state.dart';
 import 'package:c_breez/bloc/ext/block_builder_extensions.dart';
 import 'package:c_breez/bloc/lsp/lsp_bloc.dart';
-import 'package:c_breez/bloc/lsp/lsp_state.dart';
 import 'package:c_breez/routes/create_invoice/qr_code_dialog.dart';
 import 'package:c_breez/routes/create_invoice/widgets/successful_payment.dart';
 import 'package:c_breez/routes/lnurl/withdraw/lnurl_withdraw_dialog.dart';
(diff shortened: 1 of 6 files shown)


lib/routes/home/widgets/bottom_actions_bar/bottom_actions_bar.dart
```diff
@@ -26,7 +26,7 @@ class BottomActionsBar extends StatelessWidget {
     return BlocBuilder<AccountBloc, AccountState>(
       builder: (context, account) {
         return BlocBuilder<LSPBloc, LspState?>(builder: (context, lspState) {
-          final connected = lspState?.lspInfo != null;
+          final selectedLsp = (lspState?.selectedLspId != null);
```
erdemyerebasmaz, line 29: I think `isLspSelected` is a better suited name for this variable, in all instances it's used.

lib/routes/create_invoice/create_invoice_page.dart
```diff
@@ -172,13 +173,18 @@ class CreateInvoicePageState extends State<CreateInvoicePage> {
...
       barrierDismissible: false,
-      builder: (_) => LNURLWithdrawDialog(
-        requestData: data,
-        amountSats: currencyBloc.state.bitcoinCurrency.parse(
-          _amountController.text,
-        ),
-        domain: widget.domain!,
-        onFinish: widget.onFinish!,
+      builder: (_) => BlocBuilder<LSPBloc, LspState?>(
+        buildWhen: (previous, current) => previous?.lspInfo != null || current?.lspInfo != null,
```
roeierez, line 177: Why do we check previous? also what will be the exprience here

ubbabeck: Checked the previous state since if the previous states was connected I assumed that the current will also be connected. When I think about it I am not sure if this is the case and maybe the check is redundant ?

ubbabeck:
The latest commit removed checking if we have selected a LSP and also thus not showing UI as disabled. 

We can do this since the app comes with a preselected LSP so we can assume that a LSP always will be selected.

Test shows no regression as long as a LSP is selected.

lib/routes/home/widgets/bottom_actions_bar/send_options_bottom_sheet.dart
```diff
@@ -47,9 +44,9 @@ class SendOptionsBottomSheet extends StatelessWidget {
             ),
             ListTile(
               enabled: false, // TODO: back to connected when we integrate with the SDK
-              leading: BottomActionItemImage(
+              leading: const BottomActionItemImage(
                 iconAssetPath: "src/icon/bitcoin.png",
-                enabled: connected,
+                enabled: false, //  TODO: back to connected when we integrate with the SDK
```
roeierez, line 49: Why is that?

ubbabeck: I have now updated the comment to avoid confusion.
 
This was an earlier comment dependent on whenever we where connected to the LSP, where we disabled the button when not connected to the LSP.

roeierez (approved):

PR #516, opened 2023-02-27
Improve launcher icons

ademar111190:
I just saw we made new icons for c-breez, which look gorgeous and make it easier to differentiate from breez mobile; I just made some improvements saving about 120kb on the icons and adding a debug variant as we do on breez mobile to make it easier to differentiate release c-breez (blue-ish) from debug c-breez (green-ish).


roeierez (approved):
Those thing that make our life a lot easier! Thank you @ademar111190

Issue #497, opened 2023-02-21
Lightning address doesn't work 


kingonly:

erdemyerebasmaz:
LNURL's from https://lightsats.com/signin no longer work either. They did work when we were testing deep links(~3 weeks ago).

Example address:

> lnurl1dp68gurn8ghj7mrfva58gumpw3ejucm0d5hkzurf9ash2arg9akxuatjdshkgmedd3hkw6tw8a4nz0fnvcunzerzv33nzv33xq6nydp3893xyveexdnxxdtpx93rydfjvsmx2e3cx5cnyvny8ymr2etxvcunse34vcmrycfnvs6r2dtrxyuk2fn5v9nn6mr0va5kuvsme0w

We're still using dart_lnurl plugin to parse so don't bother checking SDK.

roeierez:
Let's also remove the dart_lnurl as part of this fix.

erdemyerebasmaz:
> Let's also remove the dart_lnurl as part of this fix.

There are still missing parts from the SDK, I've opened the ones I noticed. 
- https://github.com/breez/breez-sdk/issues/117 
- https://github.com/breez/breez-sdk/issues/118

Even if everything is ready, it's a long task to replace dart_lnurl logic from LNURL payment & withdraw widgets, InputHandler, InputState etc. 
I suggest we keep this restricted to lightning address & the lnurl in question for now. 
We can discuss later if dart_lnurl is the reason for this issue & using SDK will address it w/o removal of existing features.

kingonly:
@erdemyerebasmaz the question isn't if dart_lnurl is good or bad. Breez Cloud is a reference app for the Breez SDK. The Breez SDK includes lnurl functionality and Breez Cloud need to showcase how to implement lnurl using the Breez SDK only.

ademar111190:
Hey guys, sorry for the long time to give updates; my internet stopped working suddenly today.

I did start the task by writing some unit tests, as you see on this draft https://github.com/breez/c-breez/pull/507

I concluded the problem is not on c-breez, so the bug is on the input parser.

As the idea is to remove the dart_lnurl, I'll move to the SDK to continue working on the issue and use this draft PR I shared above to integrate the changes.

P.S. We should start writing more unit tests more frequently and try to convert each bug report into a unit test to avoid regressions; if we had already done it, we would have noticed the lnurl stop working earlier.

kingonly:
@ademar111190 definitely re the unit tests!

[...]

PR #527, opened 2023-03-07
Add LN Address information to Payment Details Dialog

erdemyerebasmaz:
This PR is part of 
- https://github.com/breez/c-breez/issues/497

LN Address information is added to Payment Details Dialog.

It seems like ln address information isn't added properly to payment info on SDK side or we need an additional parameter when paying to an ln address to notify SDK that ln address needs to be attached to the payment.


erdemyerebasmaz:
> It seems like ln address information isn't added properly to payment info on SDK side or we need an additional parameter when paying to an ln address to notify SDK that ln address needs to be attached to the payment.

Okay, glanced at the tx's again, I'm not entirely sure why but this behavior differs from service to service.
ln.tips tx'd do have lightning info, meanwhile wallet of satoshi ln address tx's don't.

Issue #112, opened 2022-09-06
Add Lnurl metadata to payment list item


roeierez:
We want to add the lnurl metadata to the payment list the same way we do in Breez.
Also populate the list item title in these cases.

roeierez:
@erdemyerebasmaz  Can we close this?

erdemyerebasmaz:
> #### Suggestions: (@ok300)
> * Add domain information to external payment info; If title and description are equal to each other, display domain as payment title.

@roeierez yes, it can be closed. I had this suggestion which was not discussed.

roeierez:
For now let's implement it if it is implemented in breez mobile. Otherwise let's close.

erdemyerebasmaz:
> For now let's implement it if it is implemented in breez mobile. Otherwise let's close.

No, it's the same on breez mobile for majority of instances, mainly lnurl payments & withdraws.

PR #528, opened 2023-03-07
Use LNURL metadata to extract Title, Description & Avatars for lnurl payments

erdemyerebasmaz:
This PR addresses https://github.com/breez/c-breez/issues/112

LNURL metadata is used to populate title, description & avatar fields of lnurl payments.

#### Changelist:
- identifier > plain text fields are used for payment title
- plain > long-desc text fields are used for payment description
- base64 image strings are checked for avatar images
- _Updated firebase dependencies to latest_

#### Suggestions: (@ok300)
- Add domain information to external payment info; If title and description are equal to each other, display domain as payment title.

---
Depends on
- https://github.com/breez/breez-sdk/pull/135
which will fail CI, re-run checks after it's merged.


lib/routes/home/widgets/payments_list/payment_item_avatar.dart
```diff
@@ -34,7 +40,35 @@ class PaymentItemAvatar extends StatelessWidget {
     final details = paymentItem.details.data;
     final isKeySend = (details is LnPaymentDetails) ? details.keysend : false;
     final isClosedChannelPayment = paymentItem.paymentType == PaymentType.ClosedChannel;
+    bool hasMetadataDescription = false;
+    if (details is LnPaymentDetails && details.lnurlMetadata != null) {
+      final metadataMap = {
+        for (var v in json.decode(details.lnurlMetadata!)) v[0] as String: v[1],
```
roeierez, line 46: This parsing repeats itself for some properties we extract from the medata.
I think a better approach is to parse the metadata once in the constructor and save it as a Map field.
Also if json parsing failed we should fallback to empty map.
When properties are needed (image, description) we should just access the parsed map.

roeierez (changes requested):

lib/utils/extensions/payment_extensions.dart
```diff
@@ -1,34 +1,63 @@
...
+import 'package:flutter/material.dart';
+
+final _log = FimberLog("PaymentExtensions");
+
+class PaymentExtensions {
+  final Payment _payment;
+  final BreezTranslations _texts;
+  final Map<String, dynamic> _metadataMap = {};
 
-extension PaymentExtensions on Payment {
```
roeierez, line 5: @ademar111190  I certainly agree we need to handle all these details out of the widget and thanks for initiating this! 
This complexity makes me think we need another layer for `PaymentItem` and this should be on the bloc level rather than on the UI layer.
I would go here for a slightly bigger change as I think it worth it:
At the bloc level create a `PaymentItem` model for the UI (similar to this PaymentExtensions) and emit it instead of the sdk.Payment: https://github.com/breez/c-breez/blob/main/lib/bloc/account/account_state_assembler.dart#L32

I think also all the parsing should be at the constructor initializing fields or properties and then in the UI just render the values right away.

roeierez (approved):
Fantastic! Very happy to see all the details in the bloc level.

lib/utils/extensions/payment_extensions.dart
ademar111190: @roeierez I just updated the PR with the following strategy; let me know what you think:

- The UI will use a model that I named PaymentMinutiae instead of the SDK payment; this model holds the data already formatted and ready for the UI
- The account state assembler is making the conversion, i.e., It is being made by the bloc. Still, the details of the conversion I'm keeping internally on `_PaymentMinutiaeFactory` accessed through factory `PaymentMinutiae.fromPayment` to make tests easier and avoid congestion on the account state assembler file as it will grow in the future.
- I kept the payment id on the payment minutiae in case of need to access the payment model back at some point in the flow.

Issue #267, opened 2022-12-05
Add ability to configure max routing fee (absolute or %) before paying invoice


ok300:
A malicious merchant could have his LN node reachable via a single channel with exorbitant fees.

Any invoice issued would therefore also bring in extra revenue via the (monopolistically set) routing fees to his node.

Since users usually just check the invoice amount to be paid, and not the routing fee as well (since that is negligible most of the time), they risk paying any (arbitrarily high) routing fees.

Wallets can mitigate that by probing the payment path(s) and showing the user the estimated routing fee on the payment confirmation screen.

Going a step further, a wallet could have a global setting for maximum allowable routing fee as a percentage of the invoice amount, or as a fixed amount of sats.

PR #539, opened 2023-03-15
Payment options bloc and preferences

ademar111190:
Add the bloc and the preferences that store the user options, related: https://github.com/breez/c-breez/issues/267

### DIFF
--- a/lib/bloc/payment_options/payment_options_bloc.dart
+++ b/lib/bloc/payment_options/payment_options_bloc.dart
@@ -1,0 +1,77 @@
+import 'package:c_breez/bloc/payment_options/payment_options_state.dart';
+import 'package:c_breez/utils/preferences.dart';
+import 'package:fimber/fimber.dart';
+import 'package:hydrated_bloc/hydrated_bloc.dart';
+
+final _log = FimberLog("PaymentOptionsBloc");
+
+class PaymentOptionsBloc extends Cubit<PaymentOptionsState> {
+  final Preferences _preferences;
+
+  PaymentOptionsBloc(
+    this._preferences,
+  ) : super(const PaymentOptionsState.initial()) {
+    _fetchPaymentsOverrideSettings().then((_) => _log.v("Initial settings read"));
+  }
+
+  Future<void> setOverrideFeeEnabled(bool enabled) async {
+    emit(state.copyWith(
+      overrideFeeEnabled: enabled,
+      saveEnabled: true,
+    ));
+  }
+
+  Future<void> setBaseFee(int baseFee) async {
+    emit(state.copyWith(
+      baseFee: baseFee,
+      saveEnabled: true,
+    ));
+  }
+
+  Future<void> setProportionalFee(double proportionalFee) async {
+    emit(state.copyWith(
+      proportionalFee: proportionalFee,
+      saveEnabled: true,
+    ));
+  }
+
+  Future<void> resetFees() async {
+    _log.v("Resetting payments override settings to default: enabled: $kDefaultOverrideFee, "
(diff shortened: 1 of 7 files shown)


lib/bloc/payment_options/payment_options_bloc.dart
```diff
@@ -0,0 +1,79 @@
+import 'package:c_breez/bloc/payment_options/payment_options_state.dart';
+import 'package:c_breez/utils/preferences.dart';
+import 'package:fimber/fimber.dart';
+import 'package:hydrated_bloc/hydrated_bloc.dart';
+
+final _log = FimberLog("PaymentOptionsBloc");
+
+class PaymentOptionsBloc extends Cubit<PaymentOptionsState> with HydratedMixin {
```
roeierez, line 8: I see the bloc uses HydratedMixin which means it persist the last state. But I also see we don't use the last state on startup as we load from preferences.
I agree that bloc persistency fits here but then we only need to emit on committing to the state and in our case it seems like the saving from the UI is that point.
Did you consider handling the intermediate state in the UI and expose only the save function that changes the bloc state and emits?

ademar111190: Yes, I did consider it; in fact, it is being held on the UI on breez mobile, where I did get most of the implementation, and which I want to change because being at the bloc is much easier to test,  there is nothing similar to `payment_options_bloc_test.dart` on breez mobile for instance, and we had bugs in the past on this flow there…

I think HydratedMixin can be useful in a "save instance / restore instance" flow, but I didn't test it to be honest.

ademar111190: The intermediate state being in the bloc has several advantages:
- Separation of Concerns as we keep the intermediate unsaved state at the bloc helps separate the business logic from the presentation layer, i.e., The UI layer is responsible only for the drawing, style, etc…
- Reusability: The bloc can be reused across multiple widgets and screens, which is what we are already doing, and we can expand in the future by integrating both c-breez and breezmobile. If the intermediate state is kept at the UI layer and not in the bloc, it may not be possible to use it across multiple widgets and screens; in fact, at breez mobile, where we are holding the intermediate state at the UI, in a single giant widget https://github.com/breez/breezmobile/blob/master/lib/routes/payment_options/payment_options_page.dart
- IMHO is the strongest argument here: Testing because blocs are easier to test since they are independent of the UI layer. By keeping the intermediate unsaved state at the bloc, it is easier to write unit tests for the blocs, try to do a mental exercise, check the tests at `payment_options_bloc_test.dart`, and try to write the same tests to breezmobile's payment options page, pretty hard, if possible. We had bugs in the past on this flow there, which I covered on unit tests at  `payment_options_bloc_test.dart`.
- As the intermediate state is at the bloc, we know what state our application is in at any point in time; if we keep the intermediate state at the UI, one piece of the intermediate state will be at the `OverrideFee` widget, another piece will be at the `ProportionalFeeWidget`, etc… Or we are doing everything in single widget as we are doing on breezmobile.

That is my reason for using the bloc and not the UI, so I removed the Hydrated, so now we have the preferences as the single source of truth for the saved state that should reflect on payments and bloc as the single source of truth for the intermediate state that is being edited by the user which can be saved or not.

roeierez: > The local state is in the UI, and it is almost impossible to unit test it

Flutter gives a nice way to test widgets. I admit I didn't try it but itooks promising: https://docs.flutter.dev/cookbook/testing/widget/introduction

It looks like the widgetTester let's you access to all the widgets and states in the widgets tree.

roeierez (approved):

PR #556, opened 2023-05-10
Buy Bitcoin with Moonpay

ademar111190:
Related: https://github.com/breez/breez-sdk/pull/189 and https://github.com/breez/breez-sdk/pull/208
Fixes: https://github.com/breez/c-breez/issues/560

### DIFF
--- a/lib/bloc/buy_bitcoin/moonpay/moonpay_bloc.dart
+++ b/lib/bloc/buy_bitcoin/moonpay/moonpay_bloc.dart
@@ -1,0 +1,84 @@
+import 'package:breez_sdk/breez_bridge.dart';
+import 'package:breez_sdk/bridge_generated.dart';
+import 'package:breez_translations/breez_translations_locales.dart';
+import 'package:c_breez/bloc/buy_bitcoin/moonpay/moonpay_state.dart';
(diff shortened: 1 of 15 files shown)


lib/bloc/buy_bitcoin/moonpay/moonpay_bloc.dart
```diff
@@ -0,0 +1,84 @@
...
+    emit(MoonPayState.loading());
+
+    try {
+      final url = await _breezLib.buyBitcoin(BuyBitcoinProvider.MoonPay);
+      _log.v("fetchMoonpayUrl url: $url");
+      emit(MoonPayState.urlReady(url));
+    } catch (e) {
+      _log.e("fetchMoonpayUrl error: $e");
+
+      if (e is FfiException && e.message.startsWith("Swap in progress was detected for address")) {
```
roeierez, line 33: I don't think we should check for ffiexception or specific string comes from the sdk, this tends to change and it will break the app.
I would:
1. check for inprogress first and show it if exists.
2. then try optimisticly to fetch new url and just show the error using the extractExceptionMessage

roeierez (changes requested):

roeierez (approved):

Issue #488, opened 2023-02-19
Mobile Notifications


erdemyerebasmaz:
## Tasklist
- [x] Configuring Android
- [x] Handle notifications while app is
  - [x] in the background,
  - [x] suspended.
- [x] Receive a payment when the app is in the background or terminated(on Android)
- [x] Configuring iOS
  - [ ] [Specify remote notification payload to wake up notification service app extension](https://developer.apple.com/documentation/usernotifications/modifying_content_in_newly_delivered_notifications#2942065)

## Future ideas
- [ ] Display notifications,
  - [ ] Add actions to notifications for cases that require user interaction
    - [ ] Handle user interactions through notifications,
    - [ ] Wake up device screen on Android for notifications that require user interaction

erdemyerebasmaz:
Depends on 
- [ ] https://github.com/breez/breez-sdk/issues/111

PR #557, opened 2023-05-12
Draft: Mobile Notifications (Android)

erdemyerebasmaz:
This PR addresses
- https://github.com/breez/c-breez/issues/488 's 
`Receive a payment when the app is terminated or in the background(on Android)` task.

A background message handler for silent notifications is set up to handle messages while the app is in the background or terminated. 
For received payment notifications, this message handler will schedule a background work that will create polling that checks if payment_hash from notification data exists in payment list every 10 seconds until it timeouts(after a minute).


lib/utils/breez_service_initializer.dart
```diff
@@ -0,0 +1,28 @@
...
+    await breezLib.initServices(
+      config: (await Config.instance()).sdkConfig,
+      seed: seed,
+      creds: credentials.glCreds,
+    );
+    print("Initialized Services");
+    await breezLib.startNode();
+    print("Node has started");
+  }
+  return breezLib;
```
roeierez, line 27: We should also call sync to make sure the node is scheduled on the cloud.

lib/utils/payment_hash_poller.dart
```diff
@@ -0,0 +1,58 @@
...
+
+  void start() {
+    print("Start polling for payment hash: $paymentHash every 10 seconds");
+    final injector = ServiceInjector();
+    final breezLib = injector.breezLib;
+
+    Timer? paymentReceivedTimer;
+
+    paymentReceivedTimer = Timer.periodic(
+      const Duration(seconds: 10),
```
roeierez, line 24: Perhaps every 3-5 seconds is better here. It is a cheap call and we better know as soon as the payment is competed.

lib/main.dart
```diff
@@ -105,3 +113,58 @@ void main() async {
...
+      }
+
+      switch (inputData["notification_type"]) {
+        case "payment_received":
+          pollForReceivedPayment(data, taskCompleter);
+          break;
+      }
+    }
+
+    return taskCompleter.future.timeout(
```
roeierez, line 162: We should also exit early if the payment is received.

lib/main.dart
```diff
@@ -105,3 +113,58 @@ void main() async {
...
+
+@pragma('vm:entry-point') // Mandatory if the App is obfuscated or using Flutter 3.1+
+void callbackDispatcher() {
+  const timeoutDuration = Duration(seconds: 60);
+
+  Workmanager().executeTask((task, inputData) async {
+    final taskCompleter = Completer<bool>();
+
+    if (inputData != null) {
+      var data = inputData["data"] ?? inputData;
```
roeierez, line 150: Why deserialize again? I saw a deserialization code in the previous function.

roeierez (changes requested):
Very nice to see this!
I think it makes sense to add a directory `background` and use one file under this dir for every background job we add. For this case we can call it `payment_recieved` for example.
Also we can add all the relevant code to this file (payment_hash_poller).

lib/utils/payment_hash_poller.dart
erdemyerebasmaz: I agree.
I'm also thinking of adding type & time filtering to listPayments call to reduce the list size. Which should reduce time complexity of the lookup on worst case to O(n) where n is most likely to be 1.

lib/main.dart
erdemyerebasmaz: We do complete the taskCompleter when there's a match.

lib/main.dart
erdemyerebasmaz: See the comment on L134.
> We need to parse taskName from inputData as taskName is ignored on iOS
Also, tasks unique name(payment hash) is not accessible form the executeTask afaik.

We may not need notification_type as a separate field and insert it into data. We also know the platform we're sending the notification to. 
Let's rethink this to use shared code and no extra modifications on data message.

lib/utils/payment_hash_poller.dart
erdemyerebasmaz: Done 16849692e1a649732e9cb85ed63ad85e91fb13c0
We check the payment_hash against the received payments in the last 30 minutes.

lib/main.dart
erdemyerebasmaz: > Why deserialize again? I saw a deserialization code in the previous function.

Applied changes to reflect the new payload structure in https://github.com/breez/notify/pull/2 https://github.com/breez/c-breez/commit/f25f242d21381c556d2a1a3829ca7397d5ec8adb

lib/utils/payment_hash_poller.dart
```diff
@@ -0,0 +1,60 @@
...
+    paymentReceivedTimer = Timer.periodic(
+      const Duration(seconds: 5),
+      (timer) async {
+        final List<Payment> paymentList = await breezLib.listPayments(
+          filter: PaymentTypeFilter.Received,
+          fromTimestamp: DateTime.now().subtract(const Duration(minutes: 30)).millisecondsSinceEpoch,
+        );
+        for (var payment in paymentList) {
+          final detailsData = payment.details.data;
+          if (detailsData is LnPaymentDetails && detailsData.paymentHash == paymentHash) {
```
roeierez, line 32: We should also check that it is not pending

roeierez (changes requested):

lib/utils/payment_hash_poller.dart
```diff
@@ -0,0 +1,61 @@
...
+
+  void start() {
+    const int interval = 5;
+    print("Start polling for payment hash: $paymentHash every $interval seconds");
+    final injector = ServiceInjector();
+    final breezLib = injector.breezLib;
+
+    Timer? paymentReceivedTimer;
+
+    paymentReceivedTimer = Timer.periodic(
```
roeierez, line 24: paymentReceivedTimer is never stopped.
I think we should have some upper limit on how much time we poll and then cancel the timer.

roeierez (changes requested):

lib/utils/payment_hash_poller.dart
erdemyerebasmaz: Done 91638d131cfd3fc79ec45e27fe7649f069ec2f21
Cancelled after a minute if it's available.

lib/utils/payment_hash_poller.dart
```diff
@@ -0,0 +1,63 @@
...
+            onDone(true);
+            print("Payment received! Stop polling.");
+            paymentReceivedTimer!.cancel();
+          }
+        }
+        print("Payment isn't received yet. Keep polling.");
+      },
+    );
+    // Cancel payment received timer if it's still running after a minute
+    Future.delayed(const Duration(seconds: 60), () => paymentReceivedTimer?.cancel());
```
roeierez, line 43: @erdemyerebasmaz  We have this 60 seconds limit both here and higher level of the stack.
Perhaps if we change this method (start) to receive a parameter for the timeout in seconds and also make it return a future (that throws on error or timeout) it will be more self contained.

roeierez (changes requested):
Added a suggestion to maybe simplify a bit.
Also I would consider adding both `breez_services_initializer.dart `and `payment_hash_poller.dart `under a directory called background.
I think it would make it easy to find all the staff related to background jobs there as it is going to be filled with more stuff in the future.

roeierez (approved):

PR #558, opened 2023-05-28
Embedding GL certificate

roeierez:
In order for the certificate to be embedded inside the binaries we need to use it inside the code rather than in a config file.
Here we created an app_config.dart with some place holders to be overrided during build.
In addition I simplified the app config only to the items we really need an option to configure in c-breez level. All other values should be good with the SDK defaults.


ademar111190 (approved):

erdemyerebasmaz:
@roeierez Changes look good, I saw an article that's a bit related to this topic not long ago. [The Right Way to Set Environment Variables with Compile-Time Variables](https://itnext.io/secure-your-flutter-project-the-right-way-to-set-environment-variables-with-compile-time-variables-67c3163ff9f4) 
Do you think we can benefit from using compile time variables?

roeierez:
> @roeierez Changes look good, I saw an article that's a bit related to this topic not long ago. [The Right Way to Set Environment Variables with Compile-Time Variables](https://itnext.io/secure-your-flutter-project-the-right-way-to-set-environment-variables-with-compile-time-variables-67c3163ff9f4) Do you think we can benefit from using compile time variables?

@erdemyerebasmaz  This is a really good idea and thanks for letting me know.
I changed the logic and it looks more elegant.

erdemyerebasmaz:
> > @roeierez Changes look good, I saw an article that's a bit related to this topic not long ago. [The Right Way to Set Environment Variables with Compile-Time Variables](https://itnext.io/secure-your-flutter-project-the-right-way-to-set-environment-variables-with-compile-time-variables-67c3163ff9f4) Do you think we can benefit from using compile time variables?
> 
> @erdemyerebasmaz This is a really good idea and thanks for letting me know. I changed the logic and it looks more elegant.

Looks neat! Please send instructions and config file on Slack after iOS part is also complete.

Issue #574, opened 2023-06-15
Bitcoin Transfer is not localized


ademar111190:
I think adding a kind check to `_PaymentMinutiaeFactory._title` and getting the text from the translation (already exists and it is being used on breez mobile) should fix it.

![Screenshot_20230615-110355_Breez_Cloud](https://github.com/breez/c-breez/assets/1225438/b4c12f4d-9e32-4149-abee-6447d6f6b501)

PR #575, opened 2023-06-15
Fix received payment item type title localization

ademar111190:
Fixes: https://github.com/breez/c-breez/issues/574

|Before|After|
|---|---|
|Screenshot_20230615-110355_Breez_Cloud|Screenshot_20230615-114516_Breez_Cloud|


lib/models/payment_minutiae.dart
```diff
@@ -97,6 +97,10 @@ class _PaymentMinutiaeFactory {
   }
 
   String _title() {
+    if (_payment.paymentType == PaymentType.Received) {
```
roeierez, line 100: @ademar111190  Not sure I understand here. Is it we return "Bitcoin Transfer" for every received payment?

ademar111190: True! My bad.
I changed to the second option, which is checking the text itself despite of it being a fragile option.

erdemyerebasmaz (approved):

roeierez (approved):

Issue #571, opened 2023-06-14
LNURL withdraw error


Nackoo2000:
When withdrawing the funds arrive but this message is displayed:

![image](https://github.com/breez/c-breez/assets/127208515/61d2b14a-bb9b-4956-8117-4926c6e6cce4)

roeierez:
please attach the logs.

ademar111190:
@Nackoo2000 could you add more details? Perhaps the logs too as @roeierez asked? I couldn't replicate this issue.

Nackoo2000:
I use Fountain app to create an LNURL
Scan it with CBreez or copy and paste the invoice.

Sorry for the delay of the logs.

[c-breez.logs (4).zip](https://github.com/breez/c-breez/files/11759987/c-breez.logs.4.zip)

PR #579, opened 2023-06-16
Fix lnurlw

ademar111190:
Fixes https://github.com/breez/c-breez/issues/571#issuecomment-1593286917

Our `if (result.hasError) {` check was routing the success flow and the else the error flow, so the fix is just invert the if/else.
I added more log that was useful to me understand the error during the process.


erdemyerebasmaz (approved):
Ah, good catch, sorry about that :sweat:

ademar111190:
> Ah, good catch, sorry about that 😓

who never inverted an if cast the first stone 😄

Issue #565, opened 2023-06-13
Wrong parsing of lnurl metadata


roeierez:
From the following task it seems that the lnurl metadata is a list and the code expects Map<String,dynamic>

type ‘_GrowableList<dynamic>’ is not a subtype of type ‘Map<String, dynamic>’ ::
	#6      new _PaymentMinutiaeFactory (package:c_breez/models/payment_minutiae.dart:93)
	#7      new PaymentMinutiae.fromPayment (package:c_breez/models/payment_minutiae.dart:56)
	#8      assembleAccountState.<anonymous closure> (package:c_breez/bloc/account/account_sta

PR #580, opened 2023-06-16
Fix metadata parsing

ademar111190:
Fixes https://github.com/breez/c-breez/issues/565

For some reason, I was expecting a map instead of the pattern `"[[\"text/plain\", \"lorem ipsum blah blah\"]]"`

### DIFF
--- a/lib/models/payment_minutiae.dart
+++ b/lib/models/payment_minutiae.dart
@@ -91,29 +91,34 @@ class _PaymentMinutiaeFactory {
 
   void _parseMetadata(LnPaymentDetails detailsData) {
     final metadata = detailsData.lnurlMetadata;
-    if (metadata != null && metadata.isNotEmpty) {
-      try {
-        final parsed = json.decode(metadata);
-        if (parsed is List) {
-          for (var item in parsed) {
-            if (item is List && item.length == 2) {
-              final key = item[0];
-              final value = item[1];
-              if (key is String) {
-                _metadataMap[key] = value;
-              } else {
-                _log.w("Unknown runtime type of key $key");
-              }
-            } else {
-              _log.w("Unknown runtime type of item $item");
-            }
-          }
-        } else {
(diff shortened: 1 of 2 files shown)


lib/models/payment_minutiae.dart
```diff
@@ -85,13 +85,34 @@ class _PaymentMinutiaeFactory {
...
+              }
+            } else {
+              _log.w("Unknown runtime type of item $item");
+            }
+          }
+        } else {
+          _log.w("Unknown runtime type of $parsed for $metadata");
         }
+      } catch (e) {
+        _log.w("Failed to parse metadata: $metadata", ex: e);
```
JssDWt, line 115: That looks like it will work!

If you treat the `else` clauses as upfront sanity checks, you can reduce the amount of nesting needed, and make the function a little easier to read:

```
void _parseMetadata(LnPaymentDetails detailsData) {
    final metadata = detailsData.lnurlMetadata;
    if (metadata == null || metadata.isEmpty) {
      return;
    }

    try {
      final parsed = json.decode(metadata);
      if (parsed is! List) {
        _log.w("Unknown runtime type of $parsed for $metadata");
        return;
      }

      for (var item in parsed) {
        if (item is! List || item.length != 2) {
          _log.w("Unknown runtime type of item $item");
          continue;
        }

        final key = item[0];
        final value = item[1];
        if (key is! String) {
          _log.w("Unknown runtime type of key $key");
          continue;
        }

        _metadataMap[key] = value;
      }
    } catch (e) {
      _log.w("Failed to parse metadata: $metadata", ex: e);
    }
  }
```

ademar111190: It looks much better, I'll try it and update the pr

ademar111190: It worked, thank you @JssDWt !

roeierez (approved):

Issue #570, opened 2023-06-13
Old BTC address format support


Nackoo2000:
We need to support sending to the old format of the BTC address.

![image](https://github.com/breez/c-breez/assets/127208515/bd9e5daa-bd5d-430e-bc54-fba5530319c4)

ubbabeck:
It recognises the address and successfully validates it. 
```
2023-07-10T10:16:36.192659 :: V :: BitcoinAddressTextFormField :: Calling validator for bc1qpyncvfcknjx9f7xmm0jhtj2mzcelf7ccjkenn2ejk9z8yvlf33hserkwkm ::  :: 0 ::  ::  :: 
2023-07-10T10:16:36.193189 :: V :: AccountBloc :: isValidBitcoinAddress: bc1qpyncvfcknjx9f7xmm0jhtj2mzcelf7ccjkenn2ejk9z8yvlf33hserkwkm ::  :: 0 ::  ::  :: 
2023-07-10T10:16:36.195308 :: V :: BitcoinAddressTextFormField :: Address bc1qpyncvfcknjx9f7xmm0jhtj2mzcelf7ccjkenn2ejk9z8yvlf33hserkwkm validation result ValidatorHolder{valid: true, inLock: true, locked true, hash: 634443779} ::  :: 0 ::  ::  :: 
```


when trying to broadcast the transaction it returns the following error:
```
2023-07-10T10:18:53.264574 :: D :: BreezBridge :: node-logs: DEBUG   plugin-gl-plugin-internal: Read response from JSON-RPC: {\"error\":{\"code\":301,\"message\":\"Could not afford all using all 0 available UTXOs: all short\"},\"id\":0,\"jsonrpc\":\"2.0\"} ::  :: 0 ::  ::  ::
```

Nackoo2000:
The old format is not recognised when scanning from the balance screen
I have to scan it from Send>Send to BTC Address>Scan

https://github.com/breez/c-breez/assets/127208515/eb0e22e5-b952-46b3-818d-8c0c9d755a2b

ubbabeck:
We need to make sure that sending works to all of the following addresses.

- [ ] SegWit V0 (bech32 starts with bc1q..)
- [ ] P2PKH, Legacy,  (starts with 1..)
- [ ] P2SH, Script,  (Starts with 3..)
- [ ] P2TR, Taproot, (bech32m starts with bc1p..)

ok300:
[...]

Issue #148, opened 2022-10-20
Integrate Send to on-chain


roeierez:
Depends on:
- https://github.com/breez/breez-sdk/issues/99
- https://github.com/breez/breez-sdk/issues/100

Tasks:
- [ ]  UI refactor, separate widgets for:
    - Send to BTC Address: Reverse Swap
    - Unexpected Funds: Sweep    
- [ ]  Handling scanned BTC addresses
- [ ]  Integrate [Expose txid in sweep bindings](https://github.com/breez/breez-sdk/pull/478) changes

erdemyerebasmaz:
Depends on:
- [Send to on-chain: add ReverseSwapInfo API #99](https://github.com/breez/breez-sdk/issues/99)
- [Send to on-chain: add NewReverseSwap API #100](https://github.com/breez/breez-sdk/issues/100)

PR #582, opened 2023-06-20
Send to BTC Address

ademar111190:
Some work on: https://github.com/breez/c-breez/issues/148 still need to integrate to the library to make the send works.

This commit e952f00b2fa7549e747a2d99ac5eaba107aea457 fixes the https://github.com/breez/c-breez/issues/570 issue

# How it looks like (notice at the end the sweep is failing, but the flow is working):

https://github.com/breez/c-breez/assets/1225438/64853851-41cd-4046-be00-b33c8ad9876b

### DIFF
--- a/lib/bloc/account/account_bloc.dart
+++ b/lib/bloc/account/account_bloc.dart
@@ -271,6 +271,7 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
     _log.v("validatePayment: $amount, $outgoing, $channelMinimumFee");
     var accState = state;
     if (amount > accState.maxPaymentAmount) {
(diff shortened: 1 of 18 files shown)


erdemyerebasmaz (approved):
Changes look good.
I have one comment. Can we re-use BitcoinAddressTextFormField instead of the TextFormField on SendOnchainForm or is there a difference?

ademar111190:
> Changes look good. I have one comment. Can we re-use BitcoinAddressTextFormField instead of the TextFormField on SendOnchainForm? They look awfully similar except a couple of variables.

That makes sense. I believe I can reuse the whole flow, not only the Adress Form. I'll create an issue to do it later as this change will take some time to test properly.

PR #583, opened 2023-06-20
use simple dialog instead of alertdialog breez_avatar (support for smaller screens)

ubbabeck:
# About pr

In order to support devices with smaller screen and avoid the users profilename not showing we can use `SimpleDialog` instead of `AlertDialog` 

https://github.com/breez/c-breez/assets/36157890/6522f7a2-f155-45a1-a663-22824f9f9669


erdemyerebasmaz:
@ubbabeck Definitely an improvement but does this make any visual changes on other devices? We wouldn't want a regression that affects most of our users when we're improving UI for small screen devices.

Nit: Random & Gallery buttons doesn't look good to me, can we add some vertical padding to them?

_Tagging @kingonly to review the UI._

kingonly:
I don't understand if this changes the existing UI.
As a rule if thumb, everything should look and work like Breez mobile. If users didn't complain it means we don't need to change anything.

ubbabeck:
@kingonly it is not supposed to change the UI, only make it more usable for users with smaller devices and/or using larger font and icon size. The difference if that with this dialog we get more granular control than with the other one.

Thank you for the feedback @erdemyerebasmaz I added some padding in the latest commit [af4d300](https://github.com/breez/c-breez/pull/583/commits/af4d3000b964f4d5a6d6ce2d7febb798f9ad1772) so it looks like breezmobile. 

Here is a photo of how it looks with the default screen size. 
breez_avatar_normal

Updated large font and icon size 
breez_avatar_large

kingonly:
If the UI is the same, it's fine by me.
Currently, I have more padding between the text field and the buttons.
Screenshot_2023-06-21-12-18-55-56_9a3a9060ca72a38f8a1fec1d0c1a640e

ubbabeck:
Updated the padding according to `AlertDialogs` padding setting in the latest rebase [a1ba069](https://github.com/breez/c-breez/pull/583/commits/a1ba069ae6bf0dbeeee8554794268c90ad262f76).
breez_ava

kingonly:
Great!

erdemyerebasmaz (approved):
LGTM

PR #585, opened 2023-06-27
584 integrate backup

### LINKED ISSUE #584: Integrate backup
The SDK exposes the following API for backup:

* `startBackup` - to start a backup task
* `backupStatus` - to fetch the current backup status. {backedUp: <true/false>, lastBackupTime: <time>}

The following events are emitted from the ASK:

`BackupStarted`
`BackupSucceeded`
`BackupFailed`

We need to proper reflect the status using an alert icon on the top right corner whenever the backedUp is false.
In that case we should also allow the user to start a backup as we do in Breez mobile and show the progress.

ubbabeck:
# About pr

Fixes #584 

## TODO
will need to fix UI for `BackupInProgressDialog`

### DIFF
--- a/lib/widgets/backup_in_progress_dialog.dart
+++ b/lib/widgets/backup_in_progress_dialog.dart
@@ -18,9 +18,9 @@ class _BackupInProgressDialogState extends State<BackupInProgressDialog> {
   @override
   void initState() {
     super.initState();
-    if (widget.backupState.status == BackupStatus.SUCCESS) {
+    /*if (widget.backupState.status == BackupStatus.SUCCESS) {
       _pop();
-    }
+    }*/
   }
 
   @override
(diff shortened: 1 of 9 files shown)


lib/widgets/backup_in_progress_dialog.dart
```diff
@@ -0,0 +1,42 @@
...
+  const BackupInProgressDialog(this.backupBlocStream);
+
+  @override
+  createState() => BackupInProgressDialogState();
+}
+
+class BackupInProgressDialogState extends State<BackupInProgressDialog> {
+  late StreamSubscription<BackupState?> _stateSubscription;
+  @override
+  void didChangeDependencies() {
```
roeierez, line 19: I think this dialog can be simplified by only focusing on the build function and not needing to implement the didChangeDependencies at all.
Can use BlocListener (https://pub.dev/documentation/flutter_bloc/latest/flutter_bloc/BlocListener-class.html)
To determine the right backup status and pop the widget off the stack when needed.

roeierez (changes requested):
Looks good. dropped some simplification suggestions.

roeierez (approved):

PR #586, opened 2023-06-29
Fix build context crashes by improving how we make our handlers

ademar111190:
I have been catching several crashes originating from our handlers. The major cause is the `BuildContext` being hold by the handlers.

Here I'm fixing most of these crashes by using a `Handler` structure with an init/dispose structure and getting the context when needed and checking for its nullability.


erdemyerebasmaz (approved):
LGTM

Issue #591, opened 2023-07-05
Support pasting LNURL 


kingonly:
This is broken in Breez mobile as well, but paste action/dialog should support lnurl. W/o it, it very hard to work with wallets like WoS.

PR #593, opened 2023-07-05
Support pasting LNURL

erdemyerebasmaz:
This PR addresses 
- https://github.com/breez/c-breez/issues/591

Any input that can be handled by our `InputHandler` is now supported via "**Paste Invoice or ID**" option.

If input data can't be handled(_unsupported input type, empty device clipboard_) `EnterPaymentInfoDialog` will be displayed so users can manually enter their input. 

This input will also be validated using `parseInput` **API** of **SDK** and checked against supported input types, an `"Unsupported input"` error message is shown for unsupported input types(_Bitcoin Addresses_).

#### Changelist:
- Support all input types that can be handled by our `InputHandler` "**Paste Invoice or ID**" and `EnterPaymentInfoDialog`
- Handle input on clicking "**Paste Invoice or ID**"
- Update `breez_translations` for unsupported input's error message
- Expose `parseInput` **API** through `InputBloc`
  - Remove `utils/lnurl.dart` as it's made obsolete by parsing inputs through **SDK**


ubbabeck (commented):
If a user copies the the nodeID@ipaddrress:port we do not recognise it being a node ID. 

Do we wish to handle this here in C-breez or perhaps the SDK can handle this, what do you think?

Apart from this minor issue 

Looks good to me 🥳

ubbabeck (approved):
Opened an issue in breez-sdk where we can handle this nodeID@address format issue instead https://github.com/breez/breez-sdk/issues/299

ubbabeck:
@erdemyerebasmaz after discussing with the team we decided that handling nodeURI should handle at in C-breez since this is how we do it in Breezmobile. I added this checking in the latest commit [7cfbfaf](https://github.com/breez/c-breez/pull/593/commits/7cfbfaf4e4932936de271f1ddf757c6ee590bb46).

ok300:
I added this logic in the SDK as well: https://github.com/breez/breez-sdk/pull/301

If you'd rather use that, then it's no need to handle this in the UI and you can revert the last commit here.

ubbabeck:
> I added this logic in the SDK as well: https://github.com/breez/breez-sdk/pull/301
> 
> 
> 
> If you'd rather use that, then it's no need to handle this in the UI and you can revert the last commit here.

Awesome. I will revert the latest commit. Thank you🙏

erdemyerebasmaz:
Latest changes with 222c64c317edab0810b36328cbd41bd76aa74514

_pasteTapped, which renamed to _pasteFromClipboard now only handles the latest clipboard data and EnterPaymentInfoDialog is opened if: 
- there's an error getting the clipboard data,
- clipboard data is empty,
- clipboard data can't be parsed,
- parsed clipboard data can't be handled(unsupported type),

This logic looks really ugly and needs a refactor.

Context:
> Even though this function is named paste tapped it assumes that the latest inputData on InputBloc's InputState comes from device's clipboard stream.

ubbabeck (approved):
Looks good!

Tested on android emulator. NodeURI works too 👍

lib/routes/home/home_page.dart
```diff
@@ -34,17 +34,18 @@ class HomeState extends State<Home> with AutoLockMixin, HandlerContextProvider {
   final GlobalKey firstPaymentItemKey = GlobalKey();
   final ScrollController scrollController = ScrollController();
   final handlers = <Handler>[];
+  late final InputHandler inputHandler = InputHandler(
```
ademar111190, line 37: This will break the lifecycle again.

I think `SendOptionsBottomSheet` should create its own `InputHandler` dealing with the lifecycle there.

A possibility is a duplication of event handling as there will be 2 or more InputHandlers. We should think of a mechanism to avoid that.

ademar111190 (changes requested):

lib/routes/home/home_page.dart
erdemyerebasmaz: > This will break the lifecycle again.
> 
> I think `SendOptionsBottomSheet` should create its own `InputHandler` dealing with the lifecycle there.
> 
> A possibility is a duplication of event handling as there will be 2 or more InputHandlers. We should think of a mechanism to avoid that.

#  ⚠️  Do not merge until lifecycle problem is addressed. ⚠️

erdemyerebasmaz:
> This will break the lifecycle again.
> 
> I think `SendOptionsBottomSheet` should create its own `InputHandler` dealing with the lifecycle there.
> 
> A possibility is a duplication of event handling as there will be 2 or more InputHandlers. We should think of a mechanism to avoid that.

###  ⚠️  Do not merge until lifecycle problem is addressed. ⚠️

lib/routes/home/home_page.dart
ubbabeck: Can we not add the inputhandler to the _injector.dart_? @erdemyerebasmaz

roeierez (approved):

PR #596, opened 2023-07-10
Integrate connect API

erdemyerebasmaz:
This PR addresses the changes in
- https://github.com/breez/breez-sdk/pull/303

to simplify sdk initialization steps & credential management. 

#### Changelist:
- Apply Config changes 8f5d11ae7f76542749bd1a2d91d774b8f2a78111
  - Replace GreenlightCredentials related things with NodeConfig
  - Expose GreenlightCredentials via `partnerCredentials`
- Use connect API (to register, recover & connect to a node) 9a74c24cff80800dd0510099f06f90247a39c36a
  - If mnemonic's is supplied to connect, it means register or recover flow is initiated, if it's not supplied it is being read from device storage via CredentialsManager and connection will be established to existing node
  - Merge _startNewNode & restoreNode on InitialWalkthroughPage 
  - Simplify CredentialsManager
    - Rename CredentialsManager file from credential_manager to credentials_manager
    - Removed GreenlightCredentials related logic

#### Other changes:
- Update dependencies to latest resolvable 37413d1c516f2381477bf57cfd70db2e86569ce5


lib/routes/initial_walkthrough/initial_walkthrough.dart
```diff
@@ -161,10 +162,12 @@ class InitialWalkthroughPageState extends State<InitialWalkthroughPage>
 
     final themeProvider = ThemeProvider.controllerOf(context);
     try {
-      String mnemonic = bip39.generateMnemonic(strength: 128);
-      await accountBloc.startNewNode(mnemonic: mnemonic);
+      await accountBloc.connect(mnemonic: mnemonic ?? bip39.generateMnemonic(strength: 128));
     } catch (error) {
-      _log.i("Failed to register node", ex: error);
+      _log.i("Failed to ${isRestore ? "restore" : "registe"} node", ex: error);
+      if (isRestore) {
```
roeierez, line 168: We don't need this fallback here to restore, do we?
The connects should already take that into account, if it is passed with mnemonics it restores, if it is passed with empty mnemonics it will look into the secure storage for mnemonics.

roeierez (changes requested):
One small comment, other than that LGTM

roeierez:
@erdemyerebasmaz the iOS build fails probably related to fimber log upgrade

lib/routes/initial_walkthrough/initial_walkthrough.dart
erdemyerebasmaz: We do, this flow uses the mnemonics that the user has entered manually.
If it fails, EnterMnemonicsPage is opened with the same mnemonics.

How connect on AccountBloc works is:
If mnemonics is supplied, store the mnemonics. This parameter is used to differentiate if we're establishing connection with an existing node or not.
Then later on when calling the SDK's connect API, mnemonics are restored from secure storage.

erdemyerebasmaz:
> @erdemyerebasmaz the iOS build fails probably related to fimber log upgrade

@roeierez This depends on:
- https://github.com/magillus/flutter-fimber/pull/134

Looks like it will be addressed soon but we can downgrade and move on, there aren't any critical changes.

lib/routes/initial_walkthrough/initial_walkthrough.dart
roeierez: I see now. Thanks.

roeierez (approved):
LGTM once iOS issue is resolved.

roeierez:
@erdemyerebasmaz there is a regression in startup time. Attaching a video that shows a few seconds delay until the user sees the payments list and balance.

https://github.com/breez/c-breez/assets/5394889/4c0a9042-5eaf-4a29-bc61-47679b9eb87d

roeierez:
The delay is unrelated to this PR but was in the sdk level: https://github.com/breez/breez-sdk/pull/309

Issue #604, opened 2023-07-16
Balance screen not updating


Nackoo2000:
The last two payments appear in the history but the balance did not change.
![image](https://github.com/breez/c-breez/assets/127208515/a75e4143-8aee-4433-8533-5822b890efcf)
[c-breez.logs (14).zip](https://github.com/breez/c-breez/files/12063602/c-breez.logs.14.zip)

Nackoo2000:
After closing CBreez and reopen it then it updated the balance:
![image](https://github.com/breez/c-breez/assets/127208515/f483838b-b62e-4041-8a66-e24666baf811)

erdemyerebasmaz:
Depends on
- https://github.com/breez/breez-sdk/pull/341

ubbabeck:
This issue has been resolved. Balance updated both when receiving and sending, closing.

PR #605, opened 2023-07-17
Update Balance screen 

erdemyerebasmaz:
This PR partially addresses
- #604

#### Changelist:
- Converted `WalletDashboard` and it's child widgets to `StatefulWidget's`.

#### Notes:
@roeierez Haven't debugged it yet but my initial impression is this is a asynchronous communication issue regarding calling `sync` on **SDK** side(_either on a payment sent/received flow or on receiving a `BreezEvent` on `BreezBridge`_).

### DIFF
--- a/lib/routes/home/widgets/dashboard/balance_text.dart
+++ b/lib/routes/home/widgets/dashboard/balance_text.dart
@@ -1,86 +1,66 @@
 import 'package:breez_translations/breez_translations_locales.dart';
-import 'package:c_breez/bloc/account/account_bloc.dart';
 import 'package:c_breez/bloc/account/account_state.dart';
-import 'package:c_breez/bloc/currency/currency_bloc.dart';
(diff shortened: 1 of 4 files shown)


roeierez:
@erdemyerebasmaz From my research this is not the case as the sdk only returns from payment after it syncs.
Also found an issue in greenlight regarding that and is under Christian's investigation.

erdemyerebasmaz:
@roeierez Thanks, saw your discussion now. Either way, this refactoring is a plus and is more alike to breezmobile.

- Converted WalletDashboard and it's child widgets to StatefulWidget's.
- All child widgets use Bloc states from a single source,
- Extracted fiat balance text into it's own widget.

ubbabeck (approved):
Looks good! Nice refactor 👍 
Tested receive and fiat currency flow on an android emulator, did not notice any issues.

Issue #505, opened 2023-02-22
No autocomplete on restore seed, autocomplete


ok300:
The BIP39 words are uniquely identified by the first four letters.

We can use this to simplify the UX on the Restore Seed screen: after first four letters typed in, Breez could auto-complete the word and already move the UI focus on the next input field

kingonly:
We have auto complete in Breez Mobile. This a bug, not a feature...

kingonly:
Please use the same autocomplete logic of Breez Mobile

erdemyerebasmaz:
> We have auto complete in Breez Mobile. This a bug, not a feature...
> Please use the same autocomplete logic of Breez Mobile

What he means is completing the word after typing four letters(as  BIP39 words are uniquely identified by the first four letters) without the need to clicking on the word on autocomplete result.

For example:
\> qua
> quality
> quantum
> quarter
typing the fourth letter should fill the field with the matching value.

kingonly:
Gotcha. This is nice to have, but let's prioritize the alpha release first.

PR #606, opened 2023-07-17
Autocomplete on restore form

erdemyerebasmaz:
This PR addresses #505

Applies the behavior described in https://github.com/breez/c-breez/issues/505#issuecomment-1439971950:
> Autocompleting the word after typing four letters(as BIP39 words are uniquely identified by the first four letters) without the need to click on the word of autocomplete result.
> 
> For example: > qua
> 
> > quality
> > quantum
> > quarter
> 
> typing the fourth letter should fill the field with the matching value then focus on the next input field.

### DIFF
--- a/lib/routes/initial_walkthrough/mnemonics/widgets/restore_form.dart
+++ b/lib/routes/initial_walkthrough/mnemonics/widgets/restore_form.dart
@@ -126,8 +126,11 @@ class RestoreFormPageState extends State<RestoreForm> {
 
   FutureOr<List<String>> _getSuggestions(String pattern, int itemIndex) {
     var suggestionList = WORDLIST.where((item) => item.startsWith(pattern)).toList();
-    if (pattern.length == 4 && suggestionList.isNotEmpty) {
-      _selectSuggestion(suggestionList.first, itemIndex);
+    if (suggestionList.isNotEmpty && suggestionList.length == 1) {
+      widget.textEditingControllers[itemIndex].text = suggestionList.first;
+      if (itemIndex + 1 < focusNodes.length) {
+        focusNodes[itemIndex + 1].requestFocus();
+      }
       return List.empty();
     }
     return suggestionList;


lib/routes/initial_walkthrough/mnemonics/widgets/restore_form.dart
```diff
@@ -119,8 +124,12 @@ class RestoreFormPageState extends State<RestoreForm> {
...
   }
 
-  FutureOr<List<String>> _getSuggestions(pattern) {
+  FutureOr<List<String>> _getSuggestions(String pattern, int itemIndex) {
     var suggestionList = WORDLIST.where((item) => item.startsWith(pattern)).toList();
+    if (pattern.length == 4 && suggestionList.isNotEmpty) {
+      _selectSuggestion(suggestionList.first, itemIndex);
+      return List.empty();
+    }
     return suggestionList.isNotEmpty ? suggestionList : List.empty();
```
roeierez, line 133:
```suggestion
   return suggestionList
```

roeierez (changes requested):
Other than one small suggestion, LGTM

lib/routes/initial_walkthrough/mnemonics/widgets/restore_form.dart
ubbabeck: Tested with this. Did not autocomplete or fill out the word.

ubbabeck (commented):
Tested with both your example and the suggested edit, but did not fill in the word.

erdemyerebasmaz:
> Is it still relevant @erdemyerebasmaz ?

Yes, but haven't had a chance to reevaluate after Ruben's findings. This is a low priority feature.

erdemyerebasmaz:
The autocomplete is functional but it has a lot of issues if you try to change/delete the autocompleted word at the moment. 

It works really well if you get everything right on the first try but we can't rely on that :) Postponing this.

PR #608, opened 2023-07-18
Change source target of breez-sdk from local path to git repository

erdemyerebasmaz:
This PR is part of moving `breez-sdk/sdk-flutter` to it's own git repository: [breez-sdk-flutter](https://github.com/breez/breez-sdk-flutter).

#### Other changes:
- Update dependencies to latest resolvable

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -109,6 +109,9 @@ dependency_overrides:
   intl: ^0.18.1
   path: ^1.8.3
   test_api: <0.5.0 # Requires Flutter 3.10
+  # Comment-out to work with breez-sdk from git repository
+  breez_sdk:
+    path: ../breez-sdk/libs/sdk-flutter
 
 flutter:
   uses-material-design: true
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -18,7 +18,8 @@ dependencies:
   auto_size_text: ^3.0.0
   bip39: ^1.0.6
   breez_sdk:
-    path: ../breez-sdk/libs/sdk-flutter
+    git:
+      url: https://github.com/breez/breez-sdk-flutter
   breez_translations:
     git:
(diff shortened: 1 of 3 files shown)


ubbabeck (approved):
Looks good to me. 

Tested sending and receiving as well as fiat currency flow.

pubspec.yaml
```diff
@@ -18,7 +18,8 @@ dependencies:
   auto_size_text: ^3.0.0
   bip39: ^1.0.6
   breez_sdk:
-    path: ../breez-sdk/libs/sdk-flutter
+    git:
```
roeierez, line 21: @erdemyerebasmaz  This means that if we are pushing a fix to the sdk it won't be included in the online build. Only after we push the flutter plugin.
This change Look good to me as users now don't need to build the sdk themselves but we probably need also to fix the online build to:
1.Keep building the sdk as today
2. Change the reference to the local one
3. Use the dev podspec

What do you think?

roeierez (changes requested):

pubspec.yaml
erdemyerebasmaz: @roeierez As for 1-2, to keep things simple we can add `breez_sdk` that references the local path under `dependency_overrides`
```
dependency_overrides:
# Comment-out to work with plugin on git repository
   breez_sdk:
     path: ../breez-sdk/libs/sdk-flutter
```
and when there's a need to work with plugin from git repository, we can delete it/comment it out.

I haven't checked how 3. fits in all of this yet but I suppose it's as limited as pubspec for this use case.

roeierez: Why do we need the lib_breez_sdk_core.a ? The make ios creates the pod locally and there is no change in dev experience. I don't see a reason to leave it.

PR #616, opened 2023-08-09
Allow payments within inbound liquidity when LSP does not support channel openings

erdemyerebasmaz:
Payments within inbound liquidity when LSP does not support channel openings are now allowed with this PR and a "Insufficient inbound liquidity $maxInboundLiquidity" error message is displayed when an invoice with an amount larger than inbound liquidity is attempted to be created.

Other changelist:
- Downgrade mobile_scanner to resolve warnings during Android build,
- Upgrade flutter_inappwebview as the issue that prevented this upgrade has been resolved

Todo:
- Add error message to Breez-Translations

### DIFF
--- a/lib/bloc/account/account_bloc.dart
+++ b/lib/bloc/account/account_bloc.dart
@@ -234,6 +234,9 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
           (amount > accState.maxInboundLiquidity && amount <= channelMinimumFee)) {
         throw PaymentBelowSetupFeesError(channelMinimumFee);
       }
+      if (amount > accState.maxInboundLiquidity) {
+        throw PaymentExceedLiquidityError(accState.maxInboundLiquidity);
+      }
       if (amount > accState.maxAllowedToReceive) {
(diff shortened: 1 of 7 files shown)


lib/bloc/account/account_bloc.dart
```diff
@@ -234,6 +234,9 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
           (amount > accState.maxInboundLiquidity && amount <= channelMinimumFee)) {
         throw PaymentBelowSetupFeesError(channelMinimumFee);
       }
+      if (amount > accState.maxInboundLiquidity) {
```
roeierez, line 237: @erdemyerebasmaz  This will seems like an error to me. It will throw on the normal case where fees are not null and the user is trying to receive more than the inbound liquidity. We only need to throw in that case IMO if the fees are null.

erdemyerebasmaz: @roeierez That's right, thought I put this under `else if`. 
Addressed d4d9667e01efb9fbf8454bdfb47112f45a72293a.

PR #620, opened 2023-08-16
618 visual feedback dynamic fee UI when lsp does not support opening channels

### LINKED ISSUE #618: [Visual feedback] Dynamic Fee UI when LSP does not support opening channels
Dynamic fees PR introduces new messages & UI's when the connected LSP temporarily does not support opening channels.

### When a user tries to create an invoice with an amount larger than their inbound liquidity:

`"Insufficient inbound liquidity (${inboundLiquidity})"`

I suggest we
- Display a warning box and inform that the LSP does not support opening channels temporarily,
- Replace "Receive up to:" amount with inbound liquidity if it's less than maximum depositable amount,
- If inbound liquidity is 0, disable creating an invoice until connected LSP supports opening channels

for this case.

We should also allow users to pick another LSP if  current LSP does not support opening channels(_considering the LSP list has more than one LSP_). 

---

### When a user tries to create a swap address:

A warning box, at the center of where the QR image would be, with the unedited error message from the SDK is displayed.

A more user friendly & translatable message is recommended and the UI could use another polish to better convey that there was an error in generating a swap address.

ubbabeck:
# About pr

Fixes #618 

- [x] Remove warning box when channel creation is not possible for swap.
- [x] Not let user receive more than `maxInboundLiquidity` if channel creation is unavailable.
- [x] Display warning if user tries to receive more than `maxInboundLiquidity`
- [x] If inbound liquidity is 0, disable creating an invoice until connected LSP supports opening channels
- [x] Update translations


lib/routes/create_invoice/create_invoice_page.dart
```diff
@@ -54,6 +55,7 @@ class CreateInvoicePageState extends State<CreateInvoicePage> {
   final _amountController = TextEditingController();
   final _amountFocusNode = FocusNode();
   var _doneAction = KeyboardDoneAction();
+  bool channelCreationPossible = false;
```
roeierez, line 58: It is better not to save this flag as part of the state.  Since we are using the lspbloc builder in the build method we can pass the flag directly from there.
We will also get the advantage of not having to make it optional.

lib/bloc/account/account_bloc.dart
```diff
@@ -221,6 +221,7 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
     int amount,
     bool outgoing, {
     int? channelMinimumFee,
+    bool? channelCreationPossible,
```
roeierez, line 224: Can we make it not nullable? Null here doesn't make sense.

roeierez (changes requested):

ubbabeck:
## Feedback with active channel and inbound liquidity
https://github.com/breez/c-breez/assets/36157890/2bb8e51b-7f88-46cc-b74c-31c43e1dc00f

## Feedback with no active channels, but lsp selected.

https://github.com/breez/c-breez/assets/36157890/ef9179be-5ac1-4381-b7c5-b384917b71d1

kingonly:
LGTM

lib/routes/create_invoice/create_invoice_page.dart
ubbabeck: Good point. Done

lib/bloc/account/account_bloc.dart
ubbabeck: Added

erdemyerebasmaz (approved):

PR #624, opened 2023-09-14
move action buttons in a row

### LINKED ISSUE #619: Bad UI when sending
1. Send > Paste Invoice or ID
2. Start typing in the field

The buttons "Cancel" and "Approve" should be next to each other not one on top of the other.
The Guiding text should be visible - this depends on the device not every time show up the error that is why I show 2 screenshots.

image
image

ubbabeck:
Fixes #619.


ubbabeck:
https://github.com/breez/c-breez/assets/36157890/00afd519-134c-44a9-9cee-623c26a50b38

lib/routes/home/widgets/bottom_actions_bar/enter_payment_info_dialog.dart
```diff
@@ -141,45 +141,49 @@ class EnterPaymentInfoDialogState extends State<EnterPaymentInfoDialog> {
...
+        children: [
+          SimpleDialogOption(
+            padding: const EdgeInsets.symmetric(),
+            onPressed: () => Navigator.pop(context),
+            child: Text(
+              texts.payment_info_dialog_action_cancel,
+              style: themeData.primaryTextTheme.labelLarge,
+            ),
+          ),
+          _paymentInfoController.text.isNotEmpty
```
erdemyerebasmaz, line 155: This can be written as:

```
if(_paymentInfoController.text.isNotEmpty)...[
  // Widget
],
```

ademar111190 (changes requested):
The buttons being one on top of the other is not a bug, it is how the dialog adapts itself when there is not enough available space for adding the buttons in the same row.

That is especially useful when we support multiple languages, e.g.:
- English:  "approve"/"cancel" (13 chars)
- French: "approuver"/"annuler" (16 chars)
We can have scenarios where 13 chars fit in the same row but 16 do not.

The bug reported on https://github.com/breez/c-breez/issues/619 seems to be a problem with padding, we are not padding properly so the final space for the button is not what it looks like to be

ubbabeck:
big
normal


Since latest rebase this is how it looks.

lib/routes/home/widgets/bottom_actions_bar/enter_payment_info_dialog.dart
```diff
@@ -88,7 +89,7 @@ class EnterPaymentInfoDialogState extends State<EnterPaymentInfoDialog> {
                       icon: Image(
                         image: const AssetImage("src/icon/qr_scan.png"),
                         color: themeData.primaryIconTheme.color,
-                        fit: BoxFit.contain,
+                        //fit: BoxFit.contain,
```
ademar111190, line 92: Better delete than comment.

ademar111190 (approved):

PR #625, opened 2023-09-14
bump archive package

ubbabeck:
Bumped the archive package to a patched version https://github.com/breez/c-breez/security/dependabot/2.


erdemyerebasmaz (approved):
Great find!

PR #627, opened 2023-09-15
display error when invoice is paid

ubbabeck:
Display a flushbar when there is a `FfiException` when paying an invoice.

https://github.com/breez/c-breez/assets/36157890/7280d360-c986-4459-aaef-6ce51f0ef4db


lib/widgets/payment_dialogs/processing_payment_dialog.dart
```diff
@@ -98,6 +103,10 @@ class ProcessingPaymentDialogState extends State<ProcessingPaymentDialog>
       if (widget.isLnurlPayment) {
         navigator.pop(err);
       }
+      if (err is FfiException) {
+        navigator.removeRoute(_currentRoute!);
```
erdemyerebasmaz, line 107: Wouldn't this cause a null exception if this error is received for `isLnurlPayment: true`?
We would be trying to remove a route that does not exist at this point.

This would need to be wrapped with: 
```
if(_currentRoute != null && _currentRoute!.isActive){
        navigator.removeRoute(_currentRoute!);
}
```

lib/widgets/payment_dialogs/processing_payment_dialog.dart
```diff
@@ -85,6 +89,7 @@ class ProcessingPaymentDialogState extends State<ProcessingPaymentDialog>
 
   _payAndClose() {
     final navigator = Navigator.of(context);
+    final texts = getSystemAppLocalizations();
```
erdemyerebasmaz, line 92: We can use `context.texts()` here, `getSystemAppLocalizations()` is reserved for places that has no access to `BuildContext`.
```suggestion
    final texts = context.texts();
```

erdemyerebasmaz (changes requested):
Looks good, have two suggestions.

lib/widgets/payment_dialogs/processing_payment_dialog.dart
erdemyerebasmaz: Ignore this.

lib/widgets/payment_dialogs/processing_payment_dialog.dart
ubbabeck: Added thanks

erdemyerebasmaz (approved):

PR #632, opened 2023-09-22
apply payment status

ubbabeck:
Updating to reflect sdk integration https://github.com/breez/breez-sdk/pull/463.

@erdemyerebasmaz Do you think we need to change the logic to handle the cases where `PayementStatus` is failed?


erdemyerebasmaz (approved):
LGTM!

PR #633, opened 2023-09-22
remove pasting handling

ubbabeck:
This pr removes the pasting logic of the [send_options_bottom_sheet.dart](https://github.com/breez/c-breez/compare/609-multiple-pasting?expand=1#diff-2e65547a4868cf3300a6a65a4fba5576d1ffabfbde9d4295925e68fdd6f61c9f) to mimic the behaviour of breezmobile. 

I believe this can be removed since the inputhandler handles the data if the user has something in their clipboard and in my opinion the UX is also better now. The user might want to manually write in an lighting address while having an invoice in the clipboard.

### DIFF
--- a/lib/routes/home/widgets/bottom_actions_bar/send_options_bottom_sheet.dart
+++ b/lib/routes/home/widgets/bottom_actions_bar/send_options_bottom_sheet.dart
@@ -1,16 +1,12 @@
-import 'package:breez_sdk/bridge_generated.dart';
 import 'package:breez_translations/breez_translations_locales.dart';
 import 'package:c_breez/bloc/account/account_bloc.dart';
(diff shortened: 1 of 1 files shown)


ademar111190:
@ubbabeck do you mind sharing a before/after video? because it is not clear to me what is the difference in the UX you got by these changes.

ubbabeck:
## New flow
https://github.com/breez/c-breez/assets/36157890/65dea52a-c9a2-456c-9906-c0a46a010b7a

## Old flow 

https://github.com/breez/c-breez/assets/36157890/a17acce1-e36f-4975-a478-2528a820634d

erdemyerebasmaz (approved):
LGTM!

Issue #626, opened 2023-09-14
Integrate static backup


roeierez:

erdemyerebasmaz:
Integrate
- https://github.com/breez/breez-sdk/pull/448

- [ ] Allow users to download their backup from the menu on Developers page

PR #639, opened 2023-09-27
 allow export of static backup

ubbabeck:
implements https://github.com/breez/c-breez/issues/626

## Q:
Should the file be written as bytes or string?

### DIFF
--- a/lib/bloc/account/credentials_manager.dart
+++ b/lib/bloc/account/credentials_manager.dart
@@ -1,8 +1,11 @@
 import 'dart:io';
 
+import 'package:breez_sdk/bridge_generated.dart' as sdk;
+import 'package:c_breez/services/injector.dart';
 import 'package:c_breez/services/keychain.dart';
 import 'package:logging/logging.dart';
 import 'package:path_provider/path_provider.dart';
+import 'package:c_breez/config.dart';
 
 class CredentialsManager {
   final _log = Logger("CredentialsManager");
@@ -49,5 +52,12 @@ class CredentialsManager {
     } catch (e) {
       throw e.toString();
     }
+  }
+
+  Future<sdk.StaticBackupResponse> exportStaticChannelBackup() async {
+    final breezLib = ServiceInjector().breezSDK;
+    Config config = await Config.instance();
+    String workingDir = config.sdkConfig.workingDir;
+    return await breezLib.staticBackup(request: sdk.StaticBackupRequest(workingDir: workingDir));
   }
 }
--- a/lib/routes/dev/developers_view.dart
+++ b/lib/routes/dev/developers_view.dart
@@ -1,6 +1,7 @@
 import 'dart:io';
 
 import 'package:archive/archive_io.dart';
+import 'package:breez_translations/breez_translations_locales.dart';
 import 'package:c_breez/bloc/account/account_bloc.dart';
 import 'package:c_breez/config.dart';
 import 'package:c_breez/logger.dart';
@@ -116,6 +117,7 @@ class DevelopersView extends StatelessWidget {
   }
 
   void _exportStaticBackup(BuildContext context) async {
+    final texts = getSystemAppLocalizations();
     final accBloc = context.read<AccountBloc>();
     const name = "scb.recover";
     final staticBackup = await accBloc.exportStaticChannelBackup();
@@ -134,7 +136,7 @@ class DevelopersView extends StatelessWidget {
       Share.shareXFiles([storageFile]);
     } else {
       // ignore: use_build_context_synchronously
-      showFlushbar(context, title: "Static backup data is missing");
+      showFlushbar(context, title: texts.backup_export_static_error_data_missing);
     }
   }
 }
--- a/lib/routes/dev/developers_view.dart
+++ b/lib/routes/dev/developers_view.dart
@@ -2,14 +2,17 @@ import 'dart:io';
 
 import 'package:archive/archive_io.dart';
 import 'package:c_breez/bloc/account/account_bloc.dart';
+import 'package:c_breez/config.dart';
 import 'package:c_breez/logger.dart';
 import 'package:c_breez/routes/dev/command_line_interface.dart';
 import 'package:c_breez/routes/ui_test/ui_test_page.dart';
 import 'package:c_breez/widgets/back_button.dart' as back_button;
+import 'package:c_breez/widgets/flushbar.dart';
 import 'package:c_breez/widgets/route.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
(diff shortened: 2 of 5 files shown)


lib/bloc/account/credentials_manager.dart
```diff
@@ -50,4 +52,10 @@ class CredentialsManager {
       throw e.toString();
     }
   }
+
+  Future<StaticBackupResponse> exportStaticChannelBackup() async {
+    final breezLib = ServiceInjector().breezSDK;
+    Directory workingDir = await getApplicationDocumentsDirectory();
```
roeierez, line 58: Better to take the workingDir from the config rather than recalculate it here.

lib/routes/dev/developers_view.dart
```diff
@@ -109,4 +114,26 @@ class DevelopersView extends StatelessWidget {
     final zipFile = XFile(zipFilePath);
     Share.shareXFiles([zipFile]);
   }
+
+  void _exportStaticBackup(BuildContext context) async {
+    final accBloc = context.read<AccountBloc>();
+    const name = "emergency.recover";
```
roeierez, line 120: Is this the name cln use?

lib/routes/dev/developers_view.dart
```diff
@@ -109,4 +114,26 @@ class DevelopersView extends StatelessWidget {
...
+
+      Directory workingDir = await getApplicationDocumentsDirectory();
+      String filePath = '${workingDir.path}/$name';
+      File file = File(filePath);
+      await file.writeAsBytes(emergencyList, flush: true);
+      final storageFile = XFile(filePath);
+      Share.shareXFiles([storageFile]);
+    } else {
+      // ignore: use_build_context_synchronously
+      showFlushbar(context, title: "No active channels");
```
roeierez, line 136: Maybe "`Static backup data is missing`" ?

roeierez (changes requested):

lib/routes/dev/developers_view.dart
JssDWt: Yes
https://github.com/ElementsProject/lightning/blob/df10c62508aecb7aaee0fdc4166face87663f8b9/plugins/chanbackup.c#L26

lib/bloc/account/credentials_manager.dart
ubbabeck: @roeierez  It seems to me that getting the `workingDir` from the config is doing the same call as here pluss some more. 
https://github.com/breez/c-breez/blob/f10bb948ea2c115b4362bc12d93f915f58c1c789/lib/config.dart#L125-L129.
Is it not better to use the `getApplicationDocumentsDirectory()` or am I missing something ?

roeierez: It is better to keep a single source of truth for the wirkingDir

lib/routes/dev/developers_view.dart
roeierez: I see but this, as implemented right now, is confusing because recovering from emergency file is different than recovering from the gPRc interface.
Here is how we should recover from the static backup hex data: https://docs.corelightning.org/reference/lightning-recoverchannel

If we call it emergency.recover the user would think the right way to recover is this: https://docs.corelightning.org/reference/lightning-emergencyrecover

lib/bloc/account/credentials_manager.dart
ubbabeck: Good point, added.

lib/routes/dev/developers_view.dart
ubbabeck: Added

lib/routes/dev/developers_view.dart
JssDWt: Ah, I didn't find that recoverchannel function.

ubbabeck: Renamed the file to scb.recover and also return the value as is.

lib/routes/dev/developers_view.dart
```diff
@@ -72,6 +75,8 @@ class DevelopersView extends StatelessWidget {
                 icon: Icons.share,
                 function: (_) => shareLog(),
               ),
+              Choice(
+                  title: "Export static backup", icon: Icons.charging_station, function: _exportStaticBackup)
```
ademar111190, line 79: texts.backup_export_static

lib/routes/dev/developers_view.dart
```diff
@@ -109,4 +114,27 @@ class DevelopersView extends StatelessWidget {
...
+      Config config = await Config.instance();
+      String workingDir = config.sdkConfig.workingDir;
+      String filePath = '$workingDir/$name';
+      File file = File(filePath);
+      await file.writeAsString(emergencyList, flush: true);
+      final storageFile = XFile(filePath);
+      Share.shareXFiles([storageFile]);
+    } else {
+      // ignore: use_build_context_synchronously
+      showFlushbar(context, title: "Static backup data is missing");
```
ademar111190, line 137: texts.backup_export_static_error_data_missing

ademar111190 (approved):
LGTM just one request:
Can you update the breez translations to this revision [782e52c6f272a4901ff85ecb76981085b355acd1](https://github.com/breez/Breez-Translations/commit/782e52c6f272a4901ff85ecb76981085b355acd1) and change the two "in code" texts to texts references? (I'm adding the reference id)

lib/routes/dev/developers_view.dart
ubbabeck: Thank you, but do we want this to be translated? I see that it can cause confusion in costumer support cases where users have selected other languages. We do not use translated texts for this page in [Breezmobile](https://github.com/breez/breezmobile/blob/100160f3f45ac32ea705e252f148aa39fee28168/lib/routes/dev/dev.dart#L449-L460) either. Though I am not sure if this is a deliberate design choice.  
What do you think @roeierez ?

JssDWt (approved):
utACK

lib/routes/dev/developers_view.dart
ubbabeck: added

lib/routes/dev/developers_view.dart
ademar111190: You are right @ubbabeck please disconsider what I said before.

ubbabeck:
Test
- [ ] Successfully closing a channel using `lightning-recoverchannel`(https://docs.corelightning.org/reference/lightning-recoverchannel)

Nackoo2000:
@ubbabeck I tested this from the developer's screen and the menu there displays this option.
if there is a successful download it means it works right?

ubbabeck:
> if there is a successful download it means it works right?

yes, and if the node does not have an active an error will show.

PR #640, opened 2023-10-02
Integrate Send to On-chain

erdemyerebasmaz:
This PR addresses
-  #148 

### Changelist:
- Separated Reverse Swap & Sweep pages then merged common parts under WithdrawFundsBloc and widgets later on
  - This contains renaming & file structure changes
- Introduce ReverseSwapsInBloc to fetch reverse swap list in progress
  - Poll for reverse swaps in progress on ReverseSwapPage and display the list until they are completed instead of ReverseSwapPage
- Revert UI changes to FeeChooser to make it look like it's breezmobile counterpart


lib/bloc/reverse_swap_in/reverse_swaps_in_bloc.dart
```diff
@@ -0,0 +1,45 @@
+import 'dart:async';
+
+import 'package:breez_sdk/breez_bridge.dart';
+import 'package:breez_translations/breez_translations_locales.dart';
+import 'package:c_breez/bloc/reverse_swap_in/reverse_swaps_in_state.dart';
+import 'package:c_breez/utils/exceptions.dart';
+import 'package:fimber/fimber.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
+
+class ReverseSwapsInBloc extends Cubit<ReverseSwapsInState> {
```
roeierez, line 10: NIT: why `ReverseSwapInBloc` ? Maybe just `ReverseSwapBloc` ?

lib/bloc/withdraw/withdraw_funds_bloc.dart
```diff
@@ -17,17 +17,78 @@ class WithdrawFundsBloc extends Cubit<WithdrawFundsState> {
     this._breezLib,
   ) : super(WithdrawFundsState.initial());
 
+  /* Reverse Swap */
+
+  Future<ReverseSwapInfo> sendOnchain({
```
roeierez, line 22: These functions should be in the reverse_swap bloc. The withdraw is related to the sweep when the user has on-chain balance (unexpected funds in breez mobile).

lib/bloc/withdraw/withdraw_funds_state.dart
```diff
@@ -2,24 +2,31 @@ import 'package:breez_translations/generated/breez_translations.dart';
 
 class WithdrawFundsState {
   final List<FeeOption> feeOptions;
-  final String errorMessage;
+  final String? reverseSwapErrorMessage;
```
roeierez, line 5: It looks like we are using this model for both reverse swap and withdraw funds. 
I think each one of them should have different state and different bloc, they are not the same thing.

lib/bloc/reverse_swap_in/reverse_swaps_in_bloc.dart
```diff
@@ -0,0 +1,45 @@
...
+  final _log = FimberLog("ReverseSwapsInBloc");
+  final BreezSDK _breezLib;
+  late Timer timer;
+
+  ReverseSwapsInBloc(this._breezLib) : super(ReverseSwapsInState(isLoading: true)) {
+    pollReverseSwapsInProgress();
```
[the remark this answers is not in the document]
erdemyerebasmaz: We listen to notifications on Breezmobile to trigger fetches and since we don't have these notifications on SDK I assumed they would need to be fetch periodically only when the user is on ReverseSwapPage, similar to how SwapInProgress works on Receive via BTC Address page.

lib/bloc/reverse_swap_in/reverse_swaps_in_bloc.dart
erdemyerebasmaz: I had the reverse swap related logic in`ReverseSwapBloc` before I merged them under `WithdrawFundsBloc` and the naming `SwapInBloc` was used for swaps in progress, that's why I had picked it. 

`ReverseSwapBloc` & `ReverseSwapsInProgressBloc`'s have to be separate if we need to periodically fetch in progress reverse swap list only when we're on `ReverseSwapPage`.

lib/bloc/withdraw/withdraw_funds_state.dart
erdemyerebasmaz: I had separated ReverseSwap & Sweep blocs first and then merged them later to use a singular bloc as they share fetching recommended fee & generating fee options logic but I agree the error message parts on the state are a bit confusing.

I'll separate them to avoid confusion and move the fee option logic to it's own bloc.

lib/routes/withdraw/reverse_swap/confirmation_page/reverse_swap_confirmation_page.dart
```diff
@@ -0,0 +1,118 @@
...
+
+    return Scaffold(
+      appBar: AppBar(
+        title: Text(texts.sweep_all_coins_speed),
+      ),
+      body: FutureBuilder(
+        future: _fetchFeeOptionsFuture,
+        builder: (context, snapshot) {
+          if (snapshot.error != null) {
+            return _ErrorMessage(
```
roeierez, line 58: What happens here? Do we allow to retry with a button?

lib/routes/withdraw/reverse_swap/confirmation_page/widgets/reverse_swap_button.dart
```diff
@@ -0,0 +1,63 @@
...
+
+    return SingleButtonBottomBar(
+      text: texts.sweep_all_coins_action_confirm,
+      onPressed: () => _sendOnchain(context),
+    );
+  }
+
+  Future _sendOnchain(BuildContext context) async {
+    final texts = context.texts();
+    final withdrawFundsBloc = context.read<WithdrawFundsBloc>();
```
roeierez, line 36: We need to totally separate the reverse swap from the withdraw funds, also here.
These are different things using the same name was part of the lately confusion regarding the implementation.

lib/routes/withdraw/reverse_swap/confirmation_page/widgets/reverse_swap_button.dart
```diff
@@ -0,0 +1,63 @@
...
+
+    final navigator = Navigator.of(context);
+    var loaderRoute = createLoaderRoute(context);
```
roeierez, line 39: Do we want to use the easyloading instead? Was it better from development experience?

roeierez (changes requested):

lib/main.dart
```diff
@@ -65,6 +66,9 @@ void main() async {
           BlocProvider<LSPBloc>(
             create: (BuildContext context) => LSPBloc(breezLib),
           ),
+          BlocProvider<SwapInBloc>(
```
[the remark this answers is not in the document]
erdemyerebasmaz: I'm aware it only needs to be used with SwapPage due to it's polling mechanism, it has to be an error introduced during moving files/renaming.
Removed 0e5ef917579634e5f14bb8c82e0f5b0e911bd695

lib/bloc/withdraw/withdraw_funds_bloc.dart
erdemyerebasmaz: Separeted ReverseSwap, Sweep & FeeOptions blocs 01ea0504972c88bc37c0e0185db32943e476d14a

lib/routes/withdraw/reverse_swap/confirmation_page/reverse_swap_confirmation_page.dart
erdemyerebasmaz: No, user has to go to the previous page and retry again.
This is the behavior on breezmobile but it can be improved as you suggest.

lib/routes/withdraw/reverse_swap/confirmation_page/widgets/reverse_swap_button.dart
erdemyerebasmaz: It's great when you want to display loaders without context, don't think it's needed here.
It has it's benefits and I haven't checked it internal implementation but it's a bit alarming that it was last updated 16 months ago.

You can declare a global spinner instance instead of calling createLoaderRoute() as we do here and even override it when calling show().
It requires less typing and worrying about context management, but more mental load in terms of management if you don't want your app to get stuck with a spinner if you don't allow dismiss on tap.

It's a great plugin but we haven't encountered any issues that requires contextless spinners and also it not being maintained for so long is worrisome. Let's not add it unless absolutely necessary.

lib/routes/withdraw/reverse_swap/confirmation_page/reverse_swap_confirmation_page.dart
roeierez: Let's leave it as in breez mobile.

lib/bloc/fee_options/fee_options_bloc.dart
```diff
@@ -3,31 +3,37 @@ import 'dart:async';
...
-    required String toAddress,
-    required int feeRateSatsPerVbyte,
-  }) async {
-    _log.v("Sweep to address $toAddress using $feeRateSatsPerVbyte fee vByte");
-    await _breezLib.sweep(
-      toAddress: toAddress,
-      feeRateSatsPerVbyte: feeRateSatsPerVbyte,
-    );
+  /// Lookup the most recent reverse swap pair info using the Boltz API
+  Future<ReverseSwapPairInfo> fetchReverseSwapFees({int? amountSat}) async {
```
roeierez, line 20: I think this function doesn't belong here. Maybe belongs to the reverse swap bloc?

roeierez (commented):
Looks good.

erdemyerebasmaz:
Status Update: Waiting for CI to pass before merging.

lib/bloc/reverse_swap_in/reverse_swaps_in_bloc.dart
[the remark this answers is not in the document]
erdemyerebasmaz: When we're on Send to BTC Address page and have reverse swaps in progress, we display ReverseSwapsInProgress page instead and show their TX information. Polling allows us to automatically switch back to Send to BTC Address page after those tx's complete.

PR #652, opened 2023-10-04
Handle parsed BTC addresses 

erdemyerebasmaz:
This PR addresses second part of 
- #148 

BTC addresses are now parsed and handled on InputBloc & InputHandler respectively.
Users will be redirected to Send to BTC Address page and address field will be automatically filled, as well as amount field if it's available.

### TODO:
- [ ] Ignore users own funding address
- [ ] Display label & message information if necessary. Explanation below:

@kingonly Parsed BTCAddressData can have label & message fields on Breez SDK. This is not the case on Breezmobile and this requires a product decision. Do we want to display these information on Send to BTC Address page?

Label is the label of the address - e.g. name of the receiver. Message is usually used to describe the transaction to the user, it's reasoning or a simple comment.

### DIFF
--- a/lib/bloc/input/input_bloc.dart
+++ b/lib/bloc/input/input_bloc.dart
@@ -91,7 +91,8 @@ class InputBloc extends Cubit<InputState> {
         parsedInput is InputType_LnUrlWithdraw ||
         parsedInput is InputType_LnUrlAuth ||
         parsedInput is InputType_LnUrlError ||
-        parsedInput is InputType_NodeId) {
+        parsedInput is InputType_NodeId ||
+        parsedInput is InputType_BitcoinAddress) {
       return InputState(inputData: parsedInput);
     } else {
       return InputState(isLoading: false);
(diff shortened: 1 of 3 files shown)


roeierez:
I don't think the todos are needed.

roeierez (approved):
LGTM

roeierez:
@erdemyerebasmaz is this the behavior in breez mobile?

ubbabeck:
No, in breezmobile we only handle bip21 with bolt 11 invoices. I think the rational is that we would like the user to use lighting as much as possible and not going onchain for transactions.

roeierez:
In that case I am closing this issue. We need to stick to breez mobile behavior unless discussed and decided otherwise. I am not aware of this behavior.

kingonly:
We don't need to support labels/messages in c-breez, but what I didn't understand "Ignore users own funding address".

ubbabeck:
> We don't need to support labels/messages in c-breez, but what I didn't understand "Ignore users own funding address".

@kingonly I think that the meaning was to ignore handling of the users own address if it is in the clippboard, but since we do not handle btcaddresses in clipboard this is not needed.

Issue #588, opened 2023-06-29
Some log lines seem to be truncated


JssDWt:
For example these lines suddenly end. 
Is there a limit to how long a logline can be?
Maybe we're forgetting to flush something? 

```
023-06-29T19:16:14.336647 :: D :: BreezBridge :: parse error (connection error: Socket is not connect
20
2023-06-29T18:40:11.770183 :: V :: AccountPage :: AccountPage build with 30 payments ::  :: 0 ::  ::  :: 
 "transport error", details: [], metadata: MetadataMap { headers: {} }

Caused by:
   2023-06-29T19:16:14.339287 :: E :: Bre
2023-06-29T19:16:14.340427 :: D :: BreezBridge :: client response error: not connected ::  :: 0 ::  ::  ::
```

JssDWt:
I'm pretty sure the logs are written in parallel and overlapping

``` 
2023-06-29T18:40:04.081328 :: D :: BreezBridge :: received frame=Data { stream_id: StreamId(1) } ::  :: 0 ::  ::  :: 
REDACTED,
            ),
            to_countersigner_value_sat: 247724,
            revocation_pubkey: PublicKey(
                REDACTED,
            ),
            to_broadcaster_delayed_pubkey: PublicKey(
                REDACTED,
            ),
            to_broadcaster_value_sat: 0,
            to_self_delay: 144,
            offered_htlcs: [],
            received_htlcs: [],
            feerate_per_kw: 3047,
        },
    ),
    current_counterparty_signatures: Some(
        CommitmentSignatures(
            REDACTED,
            [],
        ),
    ),
    current_counterparty_commit_info: Some(
        CommitmentInfo2 {
            is_counterparty_broadcaster: true,
            to_countersigner_pubkey: PublicKey(
                REDACTED,
            ),
            to_countersigner_value_sat: 0,
            revocation_pubkey: PublicKey(
                REDACTED,
            ),
            to_broadcaster_delayed_pubkey: PublicKey(
                REDACTED,
            ),
            to_broadcaster_value_sat: 247724,
            to_self_delay: 144,
            offered_htlcs: [],
            received_htlcs: [],
            feerate_per_kw: 3047,
        },
    ),
    previous_counterparty_commit_info: None,
    channel_closed: false,
    initial_holder_value: 0,
}
ChainState {
    current_height: 770000,
    funding_depth: 0,
    funding_double_spent_depth: 0,
    closing_depth: 0,
[...]

Issue #587, opened 2023-06-29
logs should be in UTC time


JssDWt:
Currently the times in the logs are printed in local time, It would save us quite some headaches if we log in UTC time I think.

erdemyerebasmaz:
This is currently not supported by the logging library we're using.
Depends on
- https://github.com/magillus/flutter-fimber/issues/129

ademar111190:
Another reason to move for logging, I'll take a look at this issue.

ademar111190:
Added this commit https://github.com/breez/c-breez/pull/657/commits/b4336d957850786c59e8a866022ee1d6176bd7fe to the log library change and as soon as it merges we get UTC time in iso format

PR #657, opened 2023-10-05
Change the log library from fimber to logging

ademar111190:
# Reasons to change from Fimber to Logging:
- Fimber IO has concurrency issues as reported here https://github.com/breez/c-breez/issues/588
- Fimber has not been receiving updates recently, even for bugs https://github.com/magillus/flutter-fimber/issues/135 with PR for fixing them available https://github.com/magillus/flutter-fimber/pull/137 
- Although I use (a lot) the LogCat integration offered by Fimber, most of the team does not and they are already used to the logging structure of the logging library (as we are using it on breeze mobile already)

# Drawbacks
- We lost the `SizeRollingFileTree`
  - I changed the logic to keep the last ten sessions as it seems a reasonable heuristic to me; Let me know your thoughts 
- LogCat integration as mentioned earlier

# How to test
- Everyone:
  - Goes to Developers -> three dots menu -> share logs
    - A zip file with the last 10 sessions (plus the current sessions) should be shared
- For devs:
  - Make sure the log shows on your console (debug flavors only)
 
# Please also review the SDK side of the changes
https://github.com/breez/breez-sdk/pull/505

Fixes https://github.com/breez/c-breez/issues/588
Fixes https://github.com/breez/c-breez/issues/587

### DIFF
--- a/lib/app_config.dart
+++ b/lib/app_config.dart
@@ -1,9 +1,9 @@
 import 'dart:convert';
 
 import 'package:breez_sdk/bridge_generated.dart';
-import 'package:fimber/fimber.dart';
+import 'package:logging/logging.dart';
 
-final _log = FimberLog("AppConfig");
+final _log = Logger("AppConfig");
 
 class AppConfig {
   final String apiKey = const String.fromEnvironment("API_KEY");
@@ -17,7 +17,7 @@ class AppConfig {
         deviceKey: base64.decode(glKey),
       );
     } catch (e) {
-      _log.w("Failed to parse partner credentials: $e");
+      _log.warning("Failed to parse partner credentials: $e");
       return null;
     }
   }
--- a/lib/background/background_task_handler.dart
+++ b/lib/background/background_task_handler.dart
@@ -1,10 +1,10 @@
 import 'dart:io';
 
 import 'package:c_breez/background/payment_hash_poller.dart';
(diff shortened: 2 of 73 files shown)


ubbabeck (approved):
code looks good and tested 

works with logcat and vscode dev console and easy to filter.

exporting the logs also works.

ademar111190:
* Just an addendum for clarity, we lost Logcat integrations but we still have basic functionalities working such as the output and filter messages by regex and by package

erdemyerebasmaz (approved):
LGTM!

Please re-trigger CI after breez-sdk changes are in place before merging, ty.

roeierez (commented):
Looks good!
One think I am missing is the sdk logs.
Now that the sdk logs exposes `breezLogStream` shouldn't we use it here and log the records?

ademar111190:
> Looks good! One think I am missing is the sdk logs. Now that the sdk logs exposes `breezLogStream` shouldn't we use it here and log the records?

Yes; I'm doing it here:
- https://github.com/breez/c-breez/pull/657/files#diff-e61eb31d013d12616f5532636a88cfa63631dda8f7829e5424e68542214d1608R58
- https://github.com/breez/c-breez/pull/657/files#diff-8229db30d0021c6aaf6196e316b5ca60b6d16a132c553f64380cdf512c6d506fR66-R85

PR #662, opened 2023-10-09
defer account state checking

### LINKED ISSUE #642: Missing icons on the Receive and Send menus
Reproduce:
 - Kill the app
 - Open the app
 - Click on "Receive" or "Send"
 
 Result - The menus appear with a big lag and the icons in the options do not appear
 
https://github.com/breez/c-breez/assets/127208515/ba94a7c7-dee9-4bfd-a1da-76faeb8a11a6

NOT Reproducing:
 - Kill the app
 - Open the app
 - Wait ~30 sec
 - Click on "Receive" or "Send"
 
 Result: There is no lag and icons appear.

ubbabeck:
Fixes #642 

In order to make the UX for receiving and sending a bit smoother we can defer waiting for account state until the validation of the payment and not when initialing the payment. This should make it better for users if the internet speed is slow.


ademar111190:
The pages are ready for receiving the navigation without this previous check?

roeierez (approved):
LGTM

Issue #647, opened 2023-10-04
Verifying the seed phrase


Nackoo2000:
When installing Breez Cloud and creating a new node I am not asked to verify my seed phrase.
There is no warning icon about it showing up.

kingonly:
Also after a payment? Vanilla node w/o payments isn't important.

Nackoo2000:
Yes even after payments. Tested with up to 30K in the wallet and there is no pop-up.

kingonly:
But is there a warning?

kingonly:
See: https://github.com/breez/c-breez/issues/395

Nackoo2000:
I do not see a warning. Am I missing it somewhere:

https://github.com/breez/c-breez/assets/127208515/a8795b2e-3644-404e-9372-009b9ebe2080

kingonly:
This is a regression for sure

PR #664, opened 2023-10-09
Verify seed phrase

ademar111190:
Fixes: https://github.com/breez/c-breez/issues/647

# How to test

## Flow 1 (new wallets)

- Clean install c-breez
- Start a new wallet (Let's breez! option)
- You should see a warning icon at the top right
- This warning should trigger the verify seed phrase flow

## Flow 2 (restored wallets)

- Clean install c-breez
- Restore a wallet (Restore from backup option)
- You should not see a warning icon at the top right
- (We consider the phrase verified as you just restored it)

### DIFF
--- a/lib/routes/initial_walkthrough/initial_walkthrough.dart
+++ b/lib/routes/initial_walkthrough/initial_walkthrough.dart
@@ -7,9 +7,9 @@ import 'package:c_breez/theme/theme_provider.dart' as theme;
 import 'package:c_breez/utils/exceptions.dart';
 import 'package:c_breez/widgets/flushbar.dart';
 import 'package:c_breez/widgets/loader.dart';
-import 'package:logging/logging.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
+import 'package:logging/logging.dart';
 import 'package:theme_provider/theme_provider.dart';
 
 class InitialWalkthroughPage extends StatefulWidget {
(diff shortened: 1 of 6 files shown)


lib/routes/initial_walkthrough/initial_walkthrough.dart
```diff
@@ -162,7 +162,10 @@ class InitialWalkthroughPageState extends State<InitialWalkthroughPage>
 
     final themeProvider = ThemeProvider.controllerOf(context);
     try {
-      await accountBloc.connect(mnemonic: mnemonic ?? bip39.generateMnemonic(strength: 128));
+      await accountBloc.connect(
+        mnemonic: mnemonic ?? bip39.generateMnemonic(strength: 128),
```
roeierez, line 166: How about pass null in the mnemonics in `restore` case and then we don't need the `restored` flag?

roeierez (changes requested):

lib/routes/initial_walkthrough/initial_walkthrough.dart
ademar111190: Account bloc uses the null case already: https://github.com/breez/c-breez/pull/664/files#diff-5dbe6ee35eb95a61a99689c83031840c5e21fc8f6a0f24b8b42ecbc2a87ab16bR54 when it is null, it doesn't change the current state of the validation status.
So, in order to use null instead of the boolean flag, I need to change how the bloc is started, from `if (!state.initial) connect();` to something like `if (!state.initial) init();` and adding this init method.
Let me know if you prefer this way Roei.

roeierez (approved):

PR #665, opened 2023-10-09
Trigger send to BTC address flow when camera reads a BTC address

ademar111190:
Fixes: https://github.com/breez/c-breez/issues/570

# How to test

- Generates a Bitcoin address QR code
  - Try different address formats (1... 3... bc1...)
  -  Try adding the amount and leaving it empty
- Opens c-breez and then open the camera scanner
- The app should recognize the QR code and open the Send to BTC address page
- The address and the amount should be correctly filled with the data you added on the QR code

### DIFF
--- a/lib/bloc/account/account_bloc.dart
+++ b/lib/bloc/account/account_bloc.dart
@@ -11,9 +11,9 @@ import 'package:c_breez/bloc/account/payment_filters.dart';
 import 'package:c_breez/bloc/account/payment_result.dart';
 import 'package:c_breez/config.dart';
 import 'package:connectivity_plus/connectivity_plus.dart';
-import 'package:logging/logging.dart';
 import 'package:flutter_fgbg/flutter_fgbg.dart';
 import 'package:hydrated_bloc/hydrated_bloc.dart';
+import 'package:logging/logging.dart';
 import 'package:path/path.dart' as p;
 import 'package:rxdart/rxdart.dart';
 
@@ -56,7 +56,8 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
     _listenPaymentResultEvents();
   }
 
-  // TODO: _watchAccountChanges listens to every change in the local storage and assemble a new account state accordingly
+  // TODO: _watchAccountChanges listens to every change in the local storage and assemble a new account state
+  // accordingly
   _watchAccountChanges() {
     return Rx.combineLatest3<List<sdk.Payment>, PaymentFilters, sdk.NodeState?, AccountState>(
       _breezLib.paymentsStream,
--- a/lib/bloc/input/input_bloc.dart
+++ b/lib/bloc/input/input_bloc.dart
@@ -2,6 +2,9 @@ import 'dart:async';
 
 import 'package:breez_sdk/breez_sdk.dart';
 import 'package:breez_sdk/bridge_generated.dart';
+import 'package:c_breez/bloc/input/input_data.dart';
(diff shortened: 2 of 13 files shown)


ubbabeck (approved):
Looks good to me 😃

roeierez:
@ademar111190  This will trigger the onchain page also when the bitcoin address is copied to the clipboard right?
Is that the behavior in breez mobile? Is that what we want?

ubbabeck:
> @ademar111190 This will trigger the onchain page also when the bitcoin address is copied to the clipboard right?
> Is that the behavior in breez mobile? Is that what we want?

Good catch, 

> @ademar111190 This will trigger the onchain page also when the bitcoin address is copied to the clipboard right?
> Is that the behavior in breez mobile? Is that what we want?

Good catch I forgot about this when reviewing. We only want to push the reverse_swap page when scanning btc addresses and not when it is in the clipboard.

ademar111190:
That is a good point, and it is a difference in behavior that may appear in the future as well. When X is inputed, do Y if the source is Z.
I'll think in something to have this in control.

ademar111190:
I edit the PR with the following changes:

- Add InputSource to have better control over when or not trigger things 
  - e.g. Trigger the reverse swap only if the input source was the qrcode scanner https://github.com/breez/c-breez/pull/665/files#diff-4b6025f076c9f9465907f221a017480e204b6392c91078137f95c1956f717c73R157-R159
- Solve this TODO https://github.com/breez/c-breez/pull/665/files#diff-f0972fea91bdf0761434389a1e0219a17e9bcdb662954417c566e81f09298babL59 by creating an input printer to hold the InputType data to string logic so we can have a normal log https://github.com/breez/c-breez/pull/665/files#diff-f0972fea91bdf0761434389a1e0219a17e9bcdb662954417c566e81f09298babR94 like everywhere else in the project
- Remove dynamic data and make InputState explicit
  - It makes the InputState more verbose, but it makes things explicitly in the code "Explicit is better than implicit", without null checks, depending only on type check which gives us all the help of the compiler and static analysis tools

let me know your thoughts @roeierez  @ubbabeck

ademar111190:
I don't know why the CI is not passing, the error:

```
cd cbreez
  dart format -o none --set-exit-if-changed -l 110 .
  shell: /bin/bash -e {0}
  env:
    FLUTTER_ROOT: /Users/runner/hostedtoolcache/flutter/stable-[3](https://github.com/breez/c-breez/actions/runs/6470526578/job/17566932463?pr=665#step:10:3).7.12-x6[4](https://github.com/breez/c-breez/actions/runs/6470526578/job/17566932463?pr=665#step:10:4)
    PUB_CACHE: /Users/runner/hostedtoolcache/flutter/stable-3.7.12-x64/.pub-cache
    JAVA_HOME: /Users/runner/hostedtoolcache/Java_Zulu_jdk/17.0.8-7/x64
  
Changed lib/bloc/account/account_bloc.dart
Changed lib/routes/create_invoice/create_invoice_page.dart
Formatted 336 files (2 changed) in 2.2[5](https://github.com/breez/c-breez/actions/runs/6470526578/job/17566932463?pr=665#step:10:5) seconds.
Error: Process completed with exit code 1.
```

but running locally I got no diff:

```
dart format -l 110 .
Formatted 336 files (0 changed) in 0.65 seconds.
```

I've checked account_bloc.dart and create_invoice_page.dart both are okay (I even manually edit the comments to make sure it is under 110 lines)

PR #666, opened 2023-10-09
Add fee params arg to createInvoice

### LINKED ISSUE #643: Difference in the fee for opening a channel
Reproduce: 
 - Create an invoice that requires a new channel opening
 -  Pay with another device/wallet

Result: The amount should be 2500sats and it is 2501sats. The difference is insignificant but I am not sure if there is another issue behind it.
image

ok300:
Add the cheapest known feerate as argument to `createInvoice`.

Fixes #643

### DIFF
--- a/lib/routes/create_invoice/create_invoice_page.dart
+++ b/lib/routes/create_invoice/create_invoice_page.dart
@@ -78,6 +78,10 @@ class CreateInvoicePageState extends State<CreateInvoicePage> {
   Widget build(BuildContext context) {
     final texts = context.texts();
 
+    final lspState = context.watch<LSPBloc>().state;
+    final cheapestFeeParams = lspState?.lspInfo?.openingFeeParamsList.values.first;
+    final isChannelOpeningAvailable = lspState?.isChannelOpeningAvailable ?? false;
+
     return Scaffold(
       key: _scaffoldKey,
       appBar: AppBar(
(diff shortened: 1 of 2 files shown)


lib/routes/create_invoice/create_invoice_page.dart
```diff
@@ -198,9 +198,13 @@ class CreateInvoicePageState extends State<CreateInvoicePage> {
     final accountBloc = context.read<AccountBloc>();
     final currencyBloc = context.read<CurrencyBloc>();
 
+    final lspInfo = context.read<LSPBloc>().state?.lspInfo;
+    final cheapestFeeParams = lspInfo?.openingFeeParamsList.values.first;
```
roeierez, line 202: This seems to work but I think it is better to have one stream of the LSP that wraps the build method and pass the fee params explicitly to this `_createInvoice` function. 
This fee params also should be passed to the `ReceivableBTCBox`.
This way we are sure that the same fee params that are displayed to the user are sent to the create invoice and it is trivial to understand that from the code.

roeierez (changes requested):

lib/routes/create_invoice/create_invoice_page.dart
ubbabeck: we can use context.watch which listens to changes to <Block>().state. Though we need to define it earlier in the build method so that it rebuilds if there are any changes.

roeierez (approved):
LGTM

ubbabeck (approved):
tACK 

looks good to me

PR #674, opened 2023-10-11
Doesn't show the fiat amount when the sats balance is zero

ademar111190:
Hide the fiat amount showed at home page if the sats amount is zero

<details>
<summary>Screenshots</summary>

|Before|After|
|---|---|
|Screenshot_20231011_101904-min|Screenshot_20231011_102656-min|

</details>

# How to test
- Create a new wallet or get wallet without any sats
  - It should not show the fiat value
- Using a wallet that has sats
  - It should show the fiat value


ubbabeck (approved):
utACK 👍

Issue #681, opened 2023-10-17
Remove "Fine" level logs


roeierez:
CurrentlyI see in c-breez lots of verbose logs: 

[BreezSdk] {FINE} (2023-10-12T06:08:41.329401Z) : poll
[BreezSdk] {FINE} (2023-10-12T06:08:41.329607Z) : attempting to decode a frame
[BreezSdk] {FINE} (2023-10-12T06:08:41.329668Z) : <- FramedRead::poll_next;
[BreezSdk] {FINE} (2023-10-12T06:08:41.329705Z) : -- FramedRead::poll_next;
[BreezSdk] {FINE} (2023-10-12T06:08:41.329748Z) : try_reclaim_frame;

Let's filter them.

ademar111190:
@roeierez I'm thinking about the best approach between development flow and debugging, what do you think about the following approach:

- Keeping all logs at the log file level
  - That will be useful for bug reports
  - I can improve the log filter script to allow filtering by tag and level
- Skip Fine logs on Logcat/Xcode as they usually aren't useful when working with the app.

That is what the changes look like, let me know your thoughts:
https://github.com/breez/c-breez/pull/683

roeierez:
Also we should leave debug level for now as we get useful info about the node logs and the signer

PR #683, opened 2023-10-17
Improve logging for mobile development

ademar111190:
Fixes https://github.com/breez/c-breez/issues/681

- Increase log level of app to info
- Filter fine logs from logcat/xcode

### DIFF
--- a/lib/logger.dart
+++ b/lib/logger.dart
@@ -26,7 +26,7 @@ void shareLog() async {
 
 class BreezLogger {
   BreezLogger() {
-    Logger.root.level = Level.ALL;
+    Logger.root.level = Level.INFO;
 
     if (kDebugMode) {
       Logger.root.onRecord.listen((record) {
@@ -84,9 +84,6 @@ class BreezLogger {
           break;
         case "DEBUG":
           _sdkLog.config(log.line);
-          break;
-        default:
-          _sdkLog.fine(log.line);
           break;
       }
     });
--- a/lib/bloc/account/account_bloc.dart
+++ b/lib/bloc/account/account_bloc.dart
@@ -84,7 +84,7 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
   }
 
   Future _startSdkForever() async {
-    _log.fine("starting sdk forever");
+    _log.info("starting sdk forever");
     await _startSdkOnce();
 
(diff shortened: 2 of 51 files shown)


lib/logger.dart
```diff
@@ -30,8 +30,10 @@ class BreezLogger {
     if (kDebugMode) {
       Logger.root.onRecord.listen((record) {
         // Dart analyzer doesn't understand that here we are in debug mode so we have to use kDebugMode again
-        if (kDebugMode) {
-          print(_recordToString(record));
+        if (record.level >= Level.INFO) {
```
roeierez, line 33: I think the sdk signer and node logs have debug level. We do need to include them by default.

roeierez (changes requested):
Looks good. Only a small comment regarding log verbosity.

erdemyerebasmaz (commented):
If our aim is to get rid of "Fine" level SDK logs, aka all SDK logs except ERROR, WARN, INFO, DEBUG level logs. My suggestion is not logging anything on default case on the `registerBreezSdkLog()`, or set `Logger.root.level` to `Level.CONFIG;` on BreezLogger constructor with these changes that convert certain app logs from fine to info.

lib/logger.dart
ademar111190: This filter is only applied for printing (the log file will keep all logs); Do you want to print these to log too @roeierez ?

roeierez: Is there a reason to differentiate?

ademar111190: My rational is the bug report as I mention here: https://github.com/breez/c-breez/issues/681#issuecomment-1766553792

I'll remove from the file; in future we can add back if we want.

@erdemyerebasmaz I'll keep the naming changes to do not lose it; different from the sdk logs, the app logs are very few and add a lot of information about user behavior, essential on bug reports

roeierez (approved):

lib/logger.dart
```diff
@@ -26,7 +26,7 @@ void shareLog() async {
 
 class BreezLogger {
   BreezLogger() {
-    Logger.root.level = Level.ALL;
+    Logger.root.level = Level.INFO;
```
erdemyerebasmaz, line 29: @ademar111190 Think we want to log anything equal to and above CONFIG, not INFO here.

PR #685, opened 2023-10-18
Use msat in sendPayment

roeierez:


erdemyerebasmaz (approved):
LGTM

lib/widgets/payment_dialogs/payment_request_dialog.dart
```diff
@@ -68,7 +68,7 @@ class PaymentRequestDialogState extends State<PaymentRequestDialog> {
         minHeight: minHeight,
         paymentFunc: () => context
             .read<AccountBloc>()
-            .sendPayment(widget.invoice.bolt11, widget.invoice.amountMsat == 0 ? _amountToPay! : null),
+            .sendPayment(widget.invoice.bolt11, widget.invoice.amountMsat == 0 ? _amountToPay! * 1000 : null),
```
erdemyerebasmaz, line 71: One question, should we make the sats -> msats(or keeping value as msats) conversion on `amountToPay()` of `PaymentRequestInfoDialog` instead and assume `_amountToPay` used on `sendPayment` as millisatoshies?

roeierez: _amountToPay is the UI value (which is shown in sats). That's why I wanted to keep it in sats as well.

Issue #678, opened 2023-10-15
lnurlp doesn't work if the app is closed 


kingonly:
1. Make sure the app is closed
2. Copy lnurlp link from WoS
3. Open the app

Nothing happens

ademar111190:
Didn't manage to reproduce it locally.

@kingonly can you share the device specs you are using and the logs?

ubbabeck:
Not able to reproduce on iOS using CBreez 6106.1

ademar111190:
Thanks for testing @ubbabeck 
What ios version did you test on? (that is relevant because clipboard API has changed a lot between ios versions)

kingonly:
Android OnePlus 8


https://github.com/breez/c-breez/assets/31890660/42561d15-cf7a-47ef-865a-eb009b5af762

ademar111190:
Is it Android 13 @kingonly ?
Do you mind sharing the logs?

ubbabeck:
> Thanks for testing @ubbabeck 
> 
> What ios version did you test on? (that is relevant because clipboard API has changed a lot between ios versions)

iOS 16.6.1

roeierez:
@ademar111190 is this fixed by the related PR?

ademar111190:
@roeierez  I was not able to reproduce the bug; that fix was my best effort given Roy's logs.
So I kindly ask @kingonly to test it and tell us if it is working now.

ubbabeck:
@ademar111190 i have forwarded some logs which are failing to you.

ademar111190:
@ubbabeck are you talking about https://github.com/breez/c-breez/issues/694 right?
I don't think they are related; working on that issue as well.

PR #687, opened 2023-10-19
Immediately listen inputs and move the wait for node state after event

ademar111190:
Currently, we are waiting for a node state before starting to listen to the clipboard, it can happen that a device emits the clipboard earlier than others and we end in a state where we miss such an event.

I'm changing the logic of the InputBloc initialization to start watching inputs immediately and wait for a node state once it receives an event, so we don't lose any events.

As the node stream is controlled by a behavior subject, we get the state immediately once the first item is emitted so there is no difference in behavior to what we have today; on the other hand, I believe this change will solve the issue https://github.com/breez/c-breez/issues/678 

# How to test
- Copy a lnurl to your clipboard
- Goes to c-breez app
- It should be handled
- Test with the app open and closed


roeierez (approved):
LGTM

PR #692, opened 2023-10-24
689 lnurlp and lnurlw

ubbabeck:
31efea9d2f858ac489fb701169101d0bae4c5fd5 ensures that we display the selected bitcoin denomination
 
8a9c27d15ebb710c984bd5c68f2c491027d2250b makes lnurlp min and max clickable as in breezmobile.


ademar111190 (approved):
Strange; I have the impression we had that bug fixed already some time ago.
Thanks for the fix!

PR #696, opened 2023-10-24
Make route logging logs just the route name to avoid logging sensitive data shared through route arguments

ademar111190:
We have been passing sensitive data through `RouteSettings` which was being logged.
This PR removes these data and keeps only the route name.

No QA is Needed.

P.S. Thanks @ubbabeck for noticing the problem.


ubbabeck (approved):
Looks good to me 👍

Issue #694, opened 2023-10-24
Android UI stuck on pasting a LNURL to Breez Cloud


Nackoo2000:
The issue happened when trying to paste from clipboard to Breez Cloud - Breez was closed

This was on Android 8.0.0

I sent 2 sats to this wallet from WoS after the "stuck" happened. The payment went through the WoS but never appeared in BreezC
Also, the side menu and the button are working but when trying to send it does not load any sending screen for any of the methods.

![image](https://github.com/breez/c-breez/assets/127208515/19cf0365-bd18-41b1-a83a-ee8c33e23a59)
[c-breez.logs (16).zip](https://github.com/breez/c-breez/files/13114626/c-breez.logs.16.zip)

Nackoo2000:
@ademar111190  I got unstuck by upgrading my Xperia system to the latest version but still Android 8.0.0
I copied an @ LNURL from WoS and opened BreezC - the result was the same. I will record it tomorrow.
[c-breez.logs (17).zip](https://github.com/breez/c-breez/files/13126208/c-breez.logs.17.zip)

ademar111190:
thanks @Nackoo2000 waiting for the video

ademar111190:
Thanks for the recordings and logs @Nackoo2000 with your help I've found the root of the issue;
I hope once my fix is merged you can have the app working properly again.

PR #702, opened 2023-10-26
Fix lifecycle issues with sdk connection

ademar111190:
Fixes https://github.com/breez/c-breez/issues/694

The app loses connection to the SDK once it gets in the background and it is not able to reconnect once it is back.
A way to reproduce it is by minimizing the app and interacting with other apps; you may need to open several apps or heavy apps if you are using a high-end device; once the system has enough pressure to deactivate the app you will experience the bug.

The solution I'm bringing on this PR is to disconnect the SDK once the app is paused and connect once it is started or resumed.

I've tested in several scenarios using the clipboard as the "way of checking" but you can try any other functionalities; it is not a clipboard issue at all; it is a lifecycle issue; by doing the clipboard I'm just showing it is working for https://github.com/breez/c-breez/issues/678 as well as people related issues because they look similar.

# Scenario of app closed being opened

|Different cb as previous sessions|Same cb as previous sessions|
|---|---|
|<video src="https://github.com/breez/c-breez/assets/1225438/547f568f-40a1-42c4-b245-4eea1ff5ab2f"></video>|<video src="https://github.com/breez/c-breez/assets/1225438/0c967cd1-22b7-4ce5-8cbc-3ca43963e535"></video>|

# Scenario of app exited being reopened

|Different cb as previous sessions|Same cb as previous sessions|
|---|---|
|<video src="https://github.com/breez/c-breez/assets/1225438/23d98c43-5d75-4501-b16f-899bca3db447"></video>|<video src="https://github.com/breez/c-breez/assets/1225438/4d5d3a75-e4c7-4b1a-b325-a13a82cf5453"></video>|

# Scenario of app minimized being maximized

|Different cb as previous sessions|Same cb as previous sessions|
|---|---|
|<video src="https://github.com/breez/c-breez/assets/1225438/bd507fb2-d834-4ca3-82eb-9e2e51d39960"></video>|<video src="https://github.com/breez/c-breez/assets/1225438/4ff44172-eb25-42c7-97d5-bd3f057e2a32"></video>|

# Scenario of new wallet
<video src="https://github.com/breez/c-breez/assets/1225438/de77022b-2513-4bea-bdbe-1e99756bdc1f"></video>

# Scenario of restoring wallet
|Before phrase|After phrase|
|---|---|
|<video src="https://github.com/breez/c-breez/assets/1225438/323e81d2-509b-4b0e-9b30-00f3c8c06502"></video>|<video src="https://github.com/breez/c-breez/assets/1225438/fd475214-02ec-4038-a184-ce9743467b0b"></video>|

### DIFF
--- a/lib/bloc/account/account_bloc.dart
+++ b/lib/bloc/account/account_bloc.dart
@@ -5,6 +5,7 @@ import 'package:bip39/bip39.dart' as bip39;
 import 'package:breez_sdk/breez_sdk.dart';
 import 'package:breez_sdk/bridge_generated.dart' as sdk;
 import 'package:c_breez/bloc/account/account_state.dart';
+import 'package:c_breez/bloc/account/account_state_assembler.dart';
 import 'package:c_breez/bloc/account/credentials_manager.dart';
 import 'package:c_breez/bloc/account/payment_error.dart';
 import 'package:c_breez/bloc/account/payment_filters.dart';
@@ -18,17 +19,16 @@ import 'package:logging/logging.dart';
 import 'package:path/path.dart' as p;
 import 'package:rxdart/rxdart.dart';
 
-import 'account_state_assembler.dart';
-
 const maxPaymentAmount = 4294967;
 const nodeSyncInterval = 60;
+
+final _log = Logger("AccountBloc");
 
 // AccountBloc is the business logic unit that is responsible to communicating with the lightning service
 // and reflect the node state. It is responsible for:
 // 1. Synchronizing with the node state.
 // 2. Abstracting actions exposed by the lightning service.
 class AccountBloc extends Cubit<AccountState> with HydratedMixin {
-  final _log = Logger("AccountBloc");
   static const String paymentFilterSettingsKey = "payment_filter_settings";
   static const int defaultInvoiceExpiry = Duration.secondsPerHour;
 
@@ -47,8 +47,10 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
     this._breezLib,
     this._credentialsManager,
   ) : super(AccountState.initial()) {
-    // emit on every change
-    _watchAccountChanges().listen((acc) => emit(acc));
+    _watchAccountChanges().listen((acc) {
+      _log.info("State changed: $acc");
+      emit(acc);
+    });
 
     _paymentFiltersStreamController.add(state.paymentFilters);
 
@@ -59,7 +61,7 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
 
   // TODO: _watchAccountChanges listens to every change in the local storage and assemble a new account state
   // accordingly
-  _watchAccountChanges() {
+  Stream<AccountState> _watchAccountChanges() {
     return Rx.combineLatest3<List<sdk.Payment>, PaymentFilters, sdk.NodeState?, AccountState>(
       _breezLib.paymentsStream,
       paymentFiltersStream,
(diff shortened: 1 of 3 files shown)


ubbabeck (approved):
Looks good to me. 🥳 

Could you rebase to main before merging?

roeierez:
@ademar111190 can you explain what do you mean by loosing connection to the sdk? What is the beharior? Also what prevents the app to recover now when it comes to foreground again? Did you see something in the logs?

ademar111190:
Sure @roeierez , let me see if I clarify things:

# can you explain what do you mean by losing connection to the sdk?

What I mean by losing connection to the SDK is the app stopping receiving events from the SDK after entering background and backing to foreground.

# What is the behavior?

There are two points where you can easily see the behavior, but any point in the app where we should receive events from the SDK will show the same issue.

- The home page build without payments, you can see the logs `AccountPage build with 0 payments` and the screenshot of the issue.

- The logs on the clipboard; Once the app is back from foreground with a new clipboard, you can see the logs we are handling the new input and waiting for a node state that never comes.

# Also what prevents the app to recover now when it comes to foreground again?

If I understood correctly, by now you mean our current implementation, right? If that is the case, we are not handling these states yet.

If the app works when backing from the foreground, it is because the system keeps everything running as if the app is still in the foreground; Android lifecycle is not 1/0 for background, foreground; the system keep the app in difference states such as activated and deactivated but it hides from us; on apps perspective it is foreground or not.

Android does that to improve multitask experience, but it is hard to simulate these events because the strategy used by the systems changes given version, vendor, model, battery status, etc.

My understanding is while the app is in background but in a “running” state, we keep receiving events from the SDK. If we are back in the foreground, the app will work as expected.

But if while the app is in background, the system put it in a “sleep” state, that will mess with the SDK, and once the app is back to the foreground, the SDK will not be “awakened” and we see the behavior described before.

# Did you see something in the logs?

Yes; the Account page logs stuck on “0 payments” and Input Bloc stuck at `waitForNodeState`.

# I am asking this because disconnecting on pause has its own implications (performance and races). For example on every connect the signer needs to connect to the gRPC interface which prevents the app from being ready to payment immediately.

That is true; I’m open to other solutions as well; my proposal looks natural from an app perspective, but maybe I’m missing some nuances.

Good to have in mind that stopping the SDK once we get background means the app will consume less resource once in background (battery/network/etc); that is desirable in an app.

# I would really want to understand the root cause of the issue before attempting to fix it.

To me, it looks the system is changing the status of our process, and we have no control over it or any other feedback about it except the lifecycle entry points where managing the SDK solves the issue.

Do you see (or someone in the team) any other candidate as the root cause?

roeierez (approved):
LGTM

erdemyerebasmaz:
We're debugging this @ok300 and realised `_breezLib.connect() `succeeds even if there's no internet connection.
Let's hold onto this PR until we figure out the implications on the SDK & notifications.

lib/bloc/account/account_bloc.dart
```diff
@@ -84,6 +88,11 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
     await _startSdkForever();
   }
 
+  Future<void> disconnect() {
+    _log.info("disconnect");
+    return _breezLib.disconnect();
+  }
+
```
erdemyerebasmaz, line 95: Let's remove this unused function.

ademar111190: Good catch

erdemyerebasmaz (approved):
This PR does not cause any concerns regarding notifications with the latest changes and can be merged now.

Great job 👏

roeierez:
@ademar111190 can we merge this?

Issue #697, opened 2023-10-24
Integrate latest breez-sdk changes


erdemyerebasmaz:
Following PR's are close to get merged

- [ ] https://github.com/breez/breez-sdk/pull/550
- [ ] https://github.com/breez/breez-sdk/pull/553
- [ ] https://github.com/breez/breez-sdk/pull/548
- [ ] https://github.com/breez/breez-sdk/pull/521

created an issue to keep track. You can track their progress & create individual issues as you like.

ubbabeck:
Closing this as all related issues has been resolved and implemented.

PR #705, opened 2023-10-31
refactor csv export and payment filtering

ubbabeck:
- fixes issue with multiple filters not being possible
- use only one filtering function

## Testing

- Filter payments in UI
- Export payments

### DIFF
--- a/lib/bloc/account/account_bloc.dart
+++ b/lib/bloc/account/account_bloc.dart
@@ -337,18 +337,22 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
     final nonFilteredPayments = state.payments;
     final paymentFilters = state.paymentFilters;
 
-    var filteredPayments = nonFilteredPayments.where((paymentMinutiae) {
-      final fromTimestamp = paymentFilters.fromTimestamp;
-      final toTimestamp = paymentFilters.toTimestamp;
-      final milliseconds = paymentMinutiae.paymentTime.millisecondsSinceEpoch;
-      if (fromTimestamp != null && toTimestamp != null) {
-        return fromTimestamp < milliseconds && milliseconds < toTimestamp;
-      }
-      return true;
-    }).toList();
+    var filteredPayments = nonFilteredPayments;
+    // Apply date filters, if there's any
+    if (paymentFilters.fromTimestamp != null || paymentFilters.toTimestamp != null) {
+      filteredPayments = nonFilteredPayments.where((paymentMinutiae) {
+        final fromTimestamp = paymentFilters.fromTimestamp;
+        final toTimestamp = paymentFilters.toTimestamp;
+        final milliseconds = paymentMinutiae.paymentTime.millisecondsSinceEpoch;
+        if (fromTimestamp != null && toTimestamp != null) {
+          return fromTimestamp < milliseconds && milliseconds < toTimestamp;
+        }
+        return true;
+      }).toList();
+    }
 
-    if ((paymentFilters.filters == null) ||
-        (paymentFilters.filters != null && paymentFilters.filters != sdk.PaymentTypeFilter.values)) {
+    // Apply payment type filters, if there's any
+    if (paymentFilters.filters != null && paymentFilters.filters != sdk.PaymentTypeFilter.values) {
       filteredPayments = filteredPayments.where((paymentMinutiae) {
         for (var f in paymentFilters.filters!) {
           if (f == sdk.PaymentTypeFilter.Sent && paymentMinutiae.paymentType == sdk.PaymentType.Sent) {
--- a/lib/bloc/account/account_bloc.dart
+++ b/lib/bloc/account/account_bloc.dart
@@ -10,6 +10,7 @@ import 'package:c_breez/bloc/account/payment_error.dart';
 import 'package:c_breez/bloc/account/payment_filters.dart';
 import 'package:c_breez/bloc/account/payment_result.dart';
 import 'package:c_breez/config.dart';
+import 'package:c_breez/models/payment_minutiae.dart';
 import 'package:connectivity_plus/connectivity_plus.dart';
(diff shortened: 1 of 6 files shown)


ubbabeck:
ref #697

erdemyerebasmaz (approved):

lib/bloc/account/account_bloc.dart
```diff
@@ -331,4 +332,45 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
...
+          if (f == sdk.PaymentTypeFilter.ClosedChannels &&
+              paymentMinutiae.paymentType == sdk.PaymentType.ClosedChannel) {
+            return true;
+          }
+          if (f == sdk.PaymentTypeFilter.Received &&
+              paymentMinutiae.paymentType == sdk.PaymentType.Received) {
+            return true;
+          }
+          return false;
+        }
```
erdemyerebasmaz, line 370: This logic can be improved by using
`return paymentFilters.filters!.contains(paymentMinutiae.paymentType);`

erdemyerebasmaz (changes requested):
LGTM
One suggestion regarding type filtering.

lib/bloc/account/account_bloc.dart
```diff
@@ -331,4 +332,45 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
...
+          return fromTimestamp < milliseconds && milliseconds < toTimestamp;
+        }
+        return true;
+      }).toList();
+    }
+
+    // Apply payment type filters, if there's any
+    if ((paymentFilters.filters != null && paymentFilters.filters != sdk.PaymentTypeFilter.values)) {
+      filteredPayments = filteredPayments.where((paymentMinutiae) {
+        for (var f in paymentFilters.filters!) {
```
erdemyerebasmaz, line 357: paymentFilters.filters can be extracted to a variable 
`final paymentTypeFilters = paymentFilters.filters;`

This way, the non-null assertion operator(**!**) isn't necessary inside & after (`paymentTypeFilters != null`) condition.

lib/bloc/account/account_bloc.dart
erdemyerebasmaz: I now see that the `PaymentType` & `PaymentTypeFilter` aren't the same type, we can use

`return paymentTypeFilters.any((filter) => filter.name == paymentMinutiae.paymentType.name);`

However, there's a difference between `PaymentTypeFilter.ClosedChannels` & `PaymentType.ClosedChannel` which needs to be addressed on SDK.

ubbabeck: This isn't possible since the one is of type `PaymentTypeFilter`and the other is of type `PaymentType`. I think If we want this we have to override the comparable method.

lib/bloc/account/account_bloc.dart
ubbabeck: I agree and tried, but the flutter throws an `unchecked_use_of_nullable_value`

erdemyerebasmaz: It shouldn't, make sure you replace every instance of the variable. Most IDE's refactoring tools handle this automatically.

lib/bloc/account/account_bloc.dart
erdemyerebasmaz: > This isn't possible since the one is of type `PaymentTypeFilter`and the other is of type `PaymentType`. I think If we want this we have to override the comparable method.

See the comment above. We need to rename PaymentTypeFilter.ClosedChannels to PaymentTypeFilter.ClosedChannel on the SDK and do the comparison via name of the enum.

ubbabeck: Understand, I will create a pr for this.

ubbabeck:
Depends on https://github.com/breez/breez-sdk/pull/573

lib/bloc/account/account_bloc.dart
```diff
@@ -353,22 +353,13 @@ class AccountBloc extends Cubit<AccountState> with HydratedMixin {
 
     // Apply payment type filters, if there's any
     if (paymentFilters.filters != null && paymentFilters.filters != sdk.PaymentTypeFilter.values) {
+      final filters = paymentFilters.filters!;
```
erdemyerebasmaz, line 356: I meant having this before the condition itself. i.e.
```
final paymentTypeFilters = paymentFilters.filters;
if (paymentTypeFilters != null && paymentTypeFilters != sdk.PaymentTypeFilter.values) {
  filteredPayments = filteredPayments.where((paymentMinutiae) {
    return paymentTypeFilters.any((filter) => filter.name == paymentMinutiae.paymentType.name);
  }).toList();
}
```

lib/bloc/csv_exporter.dart
```diff
@@ -94,10 +94,11 @@ class CsvExporter {
     _log.info("add filter information to path started $filePath");
     final paymentFilters = accountBloc.state.paymentFilters;
     if (paymentFilters.filters != null && paymentFilters.filters != PaymentTypeFilter.values) {
+      final filters = paymentFilters.filters!;
```
erdemyerebasmaz, line 97: https://github.com/breez/c-breez/pull/705/files/0f6bad42e620b5d304314f2c5bb612c8199251fe..3aa84e42cd42b4d0e67052f47d0f4db62644154d#r1378844786 applies here as well:

```
final paymentTypeFilters = accountBloc.state.paymentFilters.filters;
if (paymentTypeFilters != null && paymentTypeFilters != PaymentTypeFilter.values) {
  loop:
  for (var filter in paymentTypeFilters) {
  ///
```

lib/bloc/account/account_bloc.dart
ubbabeck: Ahh true, my bad

erdemyerebasmaz (approved):

PR #708, opened 2023-11-02
fix iOS build

ubbabeck:
i think this is the reason for iOS build failing


ubbabeck:
running action to se if this is the case https://github.com/breez/c-breez/actions/runs/6734941066.

ubbabeck:
In order to use `flutter_inappwebview-6.0.0-beta.25` we need to build the app using Xcode version >=14.3.1 and we currently use 14.2.

ubbabeck:
MacOS 13 uses Xcode 14.3.1 by default so I think switching to this is the way to go.

ademar111190:
> MacOS 13 uses Xcode 14.3.1 by default so I think switching to this is the way to go.

That is strange; is macos-latest solving for Sonoma? If that is the case our build is broken on Xcode 15

ubbabeck:
> That is strange; is macos-latest solving for Sonoma? If that is the case our build is broken on Xcode 15

@ademar111190  macos-latest is not what the name suggests. it's actually macos-12.

ubbabeck:
CI succeeded 🥳 
https://github.com/breez/c-breez/actions/runs/6746057798 ✅

erdemyerebasmaz (approved):

erdemyerebasmaz:
I don't know why I haven't pushed this but could be an alternative solution:

[Use XCode 14.3 with xcodebuild](https://github.com/breez/c-breez/commit/55830bb171aea64962408fae857da4d2b6c9c005)

roeierez (approved):

PR #709, opened 2023-11-08
use successAction

### LINKED ISSUE #707: add PaymentHash to lnurlp
ref https://github.com/breez/breez-sdk/issues/540 and https://github.com/breez/breez-sdk/pull/550

ubbabeck:
fixes #707

### DIFF
--- a/lib/routes/lnurl/payment/lnurl_payment_handler.dart
+++ b/lib/routes/lnurl/payment/lnurl_payment_handler.dart
@@ -63,7 +63,7 @@ Future<LNURLPageResult?> handlePayRequest(
         _log.info("LNURL payment success, action: ${result.data}");
         return LNURLPageResult(
           protocol: LnUrlProtocol.Pay,
-          successAction: result.data,
+          successAction: result.data.successAction,
         );
       } else if (result is LnUrlPayResult_EndpointError) {
         _log.info("LNURL payment failed: ${result.data.reason}");


lib/routes/lnurl/payment/lnurl_payment_handler.dart
```diff
@@ -63,7 +63,7 @@ Future<LNURLPageResult?> handlePayRequest(
         _log.info("LNURL payment success, action: ${result.data}");
         return LNURLPageResult(
           protocol: LnUrlProtocol.Pay,
-          successAction: result.data,
+          successAction: result.data.successAction,
         );
       } else if (result is LnUrlPayResult_EndpointError) {
```
erdemyerebasmaz, line 68: Do we need to handle `LnUrlPayResult_PayError` as well?

ubbabeck: No, from what I understand this will be null when an error is returned.

```                           return Ok(LnUrlPayResult::PayError {
                                data: LnUrlPayErrorData {
                                    payment_hash: invoice.payment_hash,
                                    reason: e.to_string(),
                                },
```

https://github.com/breez/breez-sdk/blob/95f5e040dfb8deb10efe3b54602b34aabbf22d6a/libs/sdk-core/src/breez_services.rs#L325-L329

ubbabeck: Sorry, I answered the wrong question. 

Agree with your latest comment with handling LnUrlPayResult_PayError too.

erdemyerebasmaz (approved):
Looks good, thank for the swift response!

Could you trigger Android & iOS builds after merging?

PR #711, opened 2023-11-09
swap page emit state when closed

ubbabeck:
When we are creating a swap and we're polling the swap address it refreshes the address every 5 seconds to see if it has been used. We emit a new state every 5 seconds. 

This works when we're in the swap page, but when the we go back to the home page close() is called, but the method might still be running and thus causing a bad state. 

To avoid this behaviour I suggest that we add a check if the bloc is closed before attempting to emit new states.


erdemyerebasmaz (changes requested):
Great find! Could you apply these changes to `RevSwapsInProgressBloc` as well?

erdemyerebasmaz (approved):

PR #712, opened 2023-11-09
Update dependencies to latest and address minor issues

erdemyerebasmaz:
#### Changelist:
- **Required:** Add missing minFinalCltvExpiryDelta to placeholder LNInvoice used on tests 89b4a6b
- Update dependencies to latest 7b903d56ee306045621108d892a78e1d1a86c3af
- Use debug signing config on debug mode ec2779519ff5ba080e8af4ddd783a50c6529b45a
  - This was set set development by mistake
  - Release versions are currently built using debug signing config (@roeierez) 
- Don't compute extract exception message when error is null 65effd18f2a350af34f97509a4d9795eb675ba2d 
  - This threw a misleading false positive exception in logs as it was being computed even if error was null
-  Increase logger level to CONFIG d6ec196ec30d614d936a3762fec44e845f71bd07
   - We were missing debug logs from the SDK

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -76,7 +76,7 @@ android {
 
     buildTypes {
         debug {
-            signingConfig signingConfigs.development
+            signingConfig signingConfigs.debug
             applicationIdSuffix '.debug'
             versionNameSuffix '-DEBUG'
             resValue "string", "app_name", "Breez Cloud - Debug"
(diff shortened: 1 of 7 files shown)


erdemyerebasmaz:
Should we also add `ios/Runner.xcodeproj/project.pbxproj` to `.gitignore`?

ubbabeck (approved):
Looks good to me

PR #714, opened 2023-11-14
Integrate send all funds

roeierez:
Also moved the `fetchReverseSwapOptions` to the reverse swap bloc. I think it belongs there.


ubbabeck (approved):
locks good to me !

erdemyerebasmaz (approved):

PR #717, opened 2023-11-15
Register device for notifications

erdemyerebasmaz:
### Changelist: 
- Use paymentByHash instead of getting filtered payment list
- Don't reinitialize log streams on resuming from a state where app's process is killed
- Mark payment_received task as completed on timeout to prevent additional retries by the OS


ok300 (approved):

lib/background/payment_hash_poller.dart
```diff
@@ -34,22 +35,17 @@ class PaymentHashPoller {
...
+        if (payment.status == PaymentStatus.Complete) {
           print("Payment received! Stop polling.");
           timer.cancel();
           print("Completed payment_received task for $paymentHash.");
           taskCompleter.complete(true);
-          return;
+          
+          // TODO Remove this delay when we have a reliable way tell if the signer is connected (Greenlight webhooks)
+          // Artificial delay to keep the signer running for longer, to ensure the preimage is sent
+          return await Future.delayed(const Duration(seconds: 10));
```
roeierez, line 48: For the delay to take effect perhaps we should add it before we complete the task?

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -41,16 +43,31 @@ class UserProfileBloc extends Cubit<UserProfileState> with HydratedMixin {
...
+      registerForNotifications(breezLib);
+    });
   }
 
-  Future registerForNotifications() async {
+  Future registerForNotifications(BreezSDK breezLib) async {
     _log.info("registerForNotifications");
     String? token = await _notifications.getToken();
     if (token != null) {
       if (token != state.profileSettings.token) {
```
roeierez, line 57: We should register anyway regardless if this is already registered since it refreshes the last_update timestamp on the server. Server may remove outdated tokens based on timestamp.

roeierez (changes requested):

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -41,26 +44,39 @@ class UserProfileBloc extends Cubit<UserProfileState> with HydratedMixin {
...
-          profileSettings: state.profileSettings.copyWith(
-            token: token,
-            userID: userID,
-            registrationRequested: true,
-          ),
-        ));
-      } else {
-        _log.info("Token is the same as before");
-      }
+      _log.info("Got a new token, registering…");
```
ok300, line 56:
```suggestion
      _log.info("Retrieved token, registering…");
```

It's not necessarily a new token.

lib/bloc/user_profile/user_profile_bloc.dart
```diff
@@ -41,26 +44,39 @@ class UserProfileBloc extends Cubit<UserProfileState> with HydratedMixin {
...
-        ));
-      } else {
-        _log.info("Token is the same as before");
-      }
+      _log.info("Retrieved token, registering…");
+      String platform = Platform.isIOS
+          ? "ios"
+          : Platform.isAndroid
+              ? "android"
+              : "";
```
erdemyerebasmaz, line 61: @roeierez @ok300 @JssDWt 
We shouldn't be registering webhooks for platforms other than `ios` & `android` right?

erdemyerebasmaz: No longer attempting to register webhooks for platforms other than iOS or Android. 37ffbc3ae07792fdc75d1b2d0722ca76ff48af7b

PR #720, opened 2023-11-16
Add dev command: stop

ok300:
Added the `stop` dev command.

Note: as the command will stop the remote note before it will be able to respond to the gRPC call, it will throw an error. See https://github.com/breez/breez-sdk/blob/89af9b4e1fb66ebb06294c709d218bf5568fab57/libs/sdk-core/src/greenlight/node_api.rs#L1362-L1368

### DIFF
--- a/lib/routes/dev/command_line_interface.dart
+++ b/lib/routes/dev/command_line_interface.dart
@@ -178,6 +178,7 @@ class _CommandLineInterfaceState extends State<CommandLineInterface> {
           case 'listPayments':
           case 'listInvoices':
           case 'closeAllChannels':
+          case 'stop':
             final command = commandArgs[0].toLowerCase();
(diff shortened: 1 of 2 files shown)


lib/routes/ui_test/ui_test_page.dart
```diff
@@ -173,6 +173,7 @@ class UITestPage extends StatelessWidget {
               title: const Text("QrCodeDialog"),
               onTap: () async {
                 LNInvoice invoice = LNInvoice(
+                  network: Network.Bitcoin,
```
erdemyerebasmaz, line 176: I'll address this on main branch, please revert this and rebase afterwards.

ok300: Done

erdemyerebasmaz (approved):

PR #725, opened 2023-11-23
Notification service app extension for iOS

erdemyerebasmaz:
This PR enhances C-Breez' capacity to accept payments while operating in the background—or not at all.

Previously, we woke C-Breez app in the background through background notifications, which allowed it to accept payments while running in the background. However, there were a number of issues with this.

Apple's intended use case for background notifications is for app’s with server-based content that changes infrequently or at irregular intervals. Background notifications are used to notify the app when new content becomes available. It wakes your app in the background and gives it time to initiate downloads and update its content so the user is shown with the updated content next time they open the app.

Since background notifications are built around this idea,
> The system treats background notifications as low priority: you can use them to refresh your app’s content, but the system doesn’t guarantee their delivery. In addition, the system may throttle the delivery of background notifications if the total number becomes excessive. The number of background notifications allowed by the system depends on current conditions, but don’t try to send more than two or three per hour.

It's very like that the notification won't get delivered if the
- device is in low power mode or currently has other CPU & Network intensive apps running
- user has received multiple payments in a short timeframe
- OS doesn't think the user will launch the app soon

This is where notification service app extension comes into play. Notification service app extension operate on push notifications _that display an alert to the user*_. This extension allows us to run tasks for ~30 seconds, modify contents of a notification before displaying* it to the user.

*: There's an entitlement, `com.apple.developer.usernotifications.filtering`, that allows a notification service extension to receive push notifications without displaying the notification to the user. So, it's possible to silence push notifications where we have failures with the task etc.


ios/Breez Notification Service Extension/BreezManager.swift
```diff
@@ -0,0 +1,63 @@
...
+    category: "BreezManager"
+)
+#else
+fileprivate var log = Logger(OSLog.disabled)
+#endif
+
+// SDK events listener
+class SDKListener: EventListener {
+    func onEvent(e: BreezEvent) {
+        print("received event ", e)
```
roeierez, line 18: Isn't it better to use the listener to detect payment instead of polling?

roeierez (changes requested):
Mainly thing is to check if we can simplify the approach with race free.

PR #726, opened 2023-11-29
Use simple approach for ios notifications handling

roeierez:
An attempt to simplify things a bit.


erdemyerebasmaz (approved):
Looks solid

PR #737, opened 2023-12-07
Notification service for Android

erdemyerebasmaz:
@roeierez Created a draft PR to make it easier to review progress.


roeierez (commented):
Skeleton looks good. I think it is ready to fill in with details.
Also I think since in contrary to iOS here we get silent push so perhaps we can raise a local notification at the end with the same message we do in iOS.

android/app/build.gradle
```diff
@@ -108,4 +115,30 @@ flutter {
...
+    implementation "androidx.work:work-rxjava2:$work_version"
+
+    // optional - GCMNetworkManager support
+    implementation "androidx.work:work-gcm:$work_version"
+
+    // optional - Test helpers
+    androidTestImplementation "androidx.work:work-testing:$work_version"
+
+    // optional - Multiprocess support
+    implementation "androidx.work:work-multiprocess:$work_version"
```
roeierez, line 140: Erdem could you elaborate why we need all these optional implementatinos?

android/app/src/main/java/com/cBreez/client/BreezFCMService.kt
```diff
@@ -0,0 +1,73 @@
...
+import com.google.firebase.messaging.FirebaseMessagingService
+import com.google.firebase.messaging.RemoteMessage
+import java.util.concurrent.TimeUnit
+
+
+class BreezFCMService : FirebaseMessagingService() {
+    val TAG = "BreezFCMService"
+
+    override fun onMessageReceived(remoteMessage: RemoteMessage) {
+        Log.d(TAG, "From: ${remoteMessage.from}")
```
roeierez, line 23: Should we call super ?

android/app/src/main/java/com/cBreez/client/BreezFCMService.kt
```diff
@@ -0,0 +1,73 @@
...
+
+        // Check if message contains a data payload.
```
roeierez, line 25: Is this logic in Breez mobile needed to still be able to handle notifications in flutter? https://github.com/breez/breezmobile/blob/master/android/app/src/main/java/com/breez/client/BreezFirebaseMessagingService.java#L48

android/app/src/main/java/com/cBreez/client/BreezSdkWorker.kt
```diff
@@ -0,0 +1,160 @@
...
+}
+
+class BreezSdkWorker(appContext: Context, workerParams: WorkerParameters) :
+    CoroutineWorker(appContext, workerParams) {
+
+    override suspend fun getForegroundInfo(): ForegroundInfo {
+        return ForegroundInfo(
+            NOTIFICATION_ID_PAYMENT_RECEIVED,
+            createNotification(
+                "Incoming payment",
```
roeierez, line 58: Is this showing during the payment? If so perhaps "Receiving payment..." ?

android/app/src/main/java/com/cBreez/client/BreezSdkWorker.kt
```diff
@@ -0,0 +1,160 @@
...
+            )
+        )
+    }
+
+    override suspend fun doWork(): Result = coroutineScope {
+        // TODO(_): Improve error handling
+        val breezSDK = connectSDK(::paymentListener)
+        val paymentHash =
+            inputData.getString("PAYMENT_HASH") ?: return@coroutineScope Result.failure()
+        val payment = breezSDK?.paymentByHash(paymentHash)
```
roeierez, line 70: We need to poll here like we do in iOS right?

android/app/src/main/java/com/cBreez/client/BreezSdkWorker.kt
```diff
@@ -0,0 +1,160 @@
...
+        val apiKey = applicationContext.getString(R.string.breezApiKey)
+
+        // Create the default config
+        val glNodeConf = GreenlightNodeConfig(null, null)
+        val nodeConf = NodeConfig.Greenlight(glNodeConf)
+        val config = defaultConfig(EnvironmentType.PRODUCTION, apiKey, nodeConf)
+        // TODO(_): Set workingDir programmatically
+        config.workingDir = "/data/user/0/com.cBreez.client.debug/app_flutter"
+        // Connect to the Breez SDK make it ready for use
+        return connect(config, seed, SDKListener(paymentListener))
```
roeierez, line 111: Perhaps it is better to use a static instance of the BlockingBreezServices? like we do in iOS?

roeierez (changes requested):
Looks good.

android/app/build.gradle
erdemyerebasmaz: We don't need all of them. They were declared on the official documentation and unused dependencies will be removed later on.

android/app/src/main/java/com/cBreez/client/BreezSdkWorker.kt
erdemyerebasmaz: Yes, this is the initial message of notification. I was referencing [iOS notification](https://github.com/breez/notify/blob/3250276b139a2084b8c5fa8d76197b338e3f5639/breezsdk/init.go#L50), we should change it there as well.

android/app/build.gradle
erdemyerebasmaz: Done [Remove unnecessary dependencies](https://github.com/breez/c-breez/pull/737/commits/e8ee75199ca39fb5f38d6d0a2b5529162a8bbf28)

android/app/src/main/java/com/cBreez/client/BreezFCMService.kt
erdemyerebasmaz: Addressed

android/app/src/main/java/com/cBreez/client/BreezSdkWorker.kt
erdemyerebasmaz: Done as part of [BreezSdkWorker second draft](https://github.com/breez/c-breez/pull/737/commits/253c8520b683d788c875ba66709a8b55f592fc57)

android/app/src/main/java/com/cBreez/client/BreezSdkWorker.kt
erdemyerebasmaz: Done as part of [BreezSdkWorker second draft](https://github.com/breez/c-breez/pull/737/commits/253c8520b683d788c875ba66709a8b55f592fc57)

android/app/src/main/java/com/cBreez/client/BreezSdkWorker.kt
```diff
@@ -0,0 +1,215 @@
...
+                NotificationManager.IMPORTANCE_LOW,
+                true
+            )
+        )
+    }
+
+    override suspend fun doWork(): Result = coroutineScope {
+        val paymentHash =
+            inputData.getString("PAYMENT_HASH") ?: return@coroutineScope Result.failure()
+        CoroutineScope(Dispatchers.Default).launch {
```
roeierez, line 81: Doesn't this mean it can run in multiple thread? We want breez SDK to be created once and this will allow multiple routines won't it?
I think on Android we can perhaps just use mutex to serialize operations?

There are two basic operations we need to serialize:
1. connect to breezSDK - which should be static instance.
2. sending notifications to paymentReceivers and removing from this list.

roeierez (changes requested):

PR #739, opened 2023-12-10
Extension logs

roeierez:
Add logs to the extension process.


.github/workflows/build-ios.yml
```diff
@@ -30,7 +30,11 @@ jobs:
         uses: actions/checkout@v4
         with:
           path: 'cbreez'
-
+      - name: Install rust
+        run: |
+          rustup set auto-self-update disable
+          rustup toolchain install stable --profile minimal
+
```
erdemyerebasmaz, line 37: Is this not needed on `Build Android` workflow?

erdemyerebasmaz (commented):
Looks good, have one question:

.github/workflows/build-ios.yml
roeierez: Added that.

erdemyerebasmaz (approved):

PR #745, opened 2023-12-19
Android notifications logging

erdemyerebasmaz:
This PR adds logging capability to notification service on Android.

### Other changes:
Notifications for received payments now click app when clicked and are dismissable.

### DIFF
--- a/android/app/src/main/java/com/cBreez/client/BreezFCMService.kt
+++ b/android/app/src/main/java/com/cBreez/client/BreezFCMService.kt
@@ -8,7 +8,9 @@ import android.os.SystemClock
 import com.google.android.gms.common.util.PlatformVersion.isAtLeastLollipop
 import com.google.firebase.messaging.FirebaseMessagingService
 import com.google.firebase.messaging.RemoteMessage
+import io.flutter.util.PathUtils
 import org.tinylog.kotlin.Logger
+import java.io.File
 
 @SuppressLint("MissingFirebaseInstanceTokenRefresh")
 class BreezFcmService : FirebaseMessagingService() {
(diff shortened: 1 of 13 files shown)


android/app/src/main/kotlin/com/cBreez/client/MainActivity.kt
```diff
@@ -1,6 +1,27 @@
...
 import io.flutter.embedding.android.FlutterActivity
+import io.flutter.util.PathUtils
+import org.tinylog.kotlin.Logger
+import java.io.File
 
-class MainActivity: FlutterActivity() {
+class MainActivity : FlutterActivity() {
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        configureLogger()
```
roeierez, line 12: Is the Activity created when the app is killed and the notification is received?

ios/Breez Notification Service Extension/NotificationService.swift
```diff
@@ -9,7 +9,7 @@ class NotificationService: UNNotificationServiceExtension {
     private var logger: XCGLogger = {
         let logsDir = FileManager
             .default.containerURL(forSecurityApplicationGroupIdentifier: "group.F7R2LZH3W5.com.cBreez.client")!.appendingPathComponent("logs")
-        let extensionLogFile = logsDir.appendingPathComponent("extension-\(Date().timeIntervalSince1970).log")
+        let extensionLogFile = logsDir.appendingPathComponent("ios-extension-\(Date().timeIntervalSince1970).log")
```
erdemyerebasmaz, line 12: TODO: Add ios extension logs under `logs/ios-extension/` directory and rename file name to timestamp

android/app/src/main/java/com/cBreez/client/BreezFCMService.kt
```diff
@@ -19,25 +20,31 @@ class BreezFcmService : FirebaseMessagingService() {
 
     override fun onMessageReceived(remoteMessage: RemoteMessage) {
         super.onMessageReceived(remoteMessage)
+        configureLogger()
```
roeierez, line 23: This creates a logger and configure it with every notification. It is better IMO to configure it once and set it statically.

roeierez (changes requested):

roeierez (approved):

PR #748, opened 2023-12-26
Receive Firebase notifications when app is killed

### LINKED ISSUE #747: Handle background notifications after app is killed on Android devices
Basically, we need to

- [ ] Enable autostart
- [ ] Disable battery optimizations

for the app, but it's a bit of a complicated process as there's too many variance from manufacturer to manufacturer and even from device to device of a certain manufacturer...

erdemyerebasmaz:
Closes #747.

Edit:
Reverted deprecated WakefullBroadcastReceiver changes and added notification runtime permissions.


android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
```diff
@@ -0,0 +1,137 @@
...
+        }
+        return START_NOT_STICKY
+    }
+
+    private fun handleNow(remoteMessage: RemoteMessage): Boolean {
+        return if (remoteMessage.data["notification_type"] == "payment_received") {
+            val paymentHash = remoteMessage.data["payment_hash"]
+            val clickAction = remoteMessage.data["click_action"]
+            paymentHash?.let {
+                JobManager.instance.startPaymentReceivedJob(
```
roeierez, line 114: I don't think we need a job here as we already running in a foreground service, we should have cpu and time.

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
```diff
@@ -0,0 +1,137 @@
...
+    override fun onCreate() {
+        super.onCreate()
+        Logger.tag(TAG).debug("Creating Breez node service...")
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            registerNotificationChannels(applicationContext)
+        }
+        startForeground(
+            NOTIFICATION_ID_FOREGROUND_SERVICE,
+            notifyForegroundService(applicationContext)
+        )
```
roeierez, line 44: We already start the foreground service in onStartCommand . It seems wrong that we need to do it here as well.

android/app/src/main/java/com/cBreez/client/BreezNotificationService.kt
```diff
@@ -16,91 +26,182 @@ import kotlinx.coroutines.launch
 import org.tinylog.kotlin.Logger
 
 class BreezNotificationService {
+
```
roeierez, line 29: `BreezNotificationService` name implies Android service.... perhaps we can rename it to `BreezNotificationHelper` ?

roeierez (changes requested):

android/app/src/main/java/com/cBreez/client/BreezNotificationService.kt
erdemyerebasmaz: Done f57249306cf4e1180de1a060284a73716d50519c

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
erdemyerebasmaz: Done, but removed from onStartCommand instead c5d1e7efd906d0d756e38726b6278a726129f74e

When I removed this from onCreate, the foreground service notification persisted, dismissing programmatically after task ended didn't made any changes as well.

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
erdemyerebasmaz: [JobManager](https://github.com/breez/c-breez/blob/main/android/app/src/main/java/com/cBreez/client/JobManager.kt
) is just a proxy to enqueue work requests.

From Android 14 behavior changes: [Foreground service types are required](https://developer.android.com/about/versions/14/changes/fgs-types-required)

> If a use case in your app isn't associated with any of these types, we strongly recommend that you migrate your logic to use [WorkManager](https://developer.android.com/topic/libraries/architecture/workmanager) or [user-initiated data transfer jobs](https://developer.android.com/about/versions/14/changes/user-initiated-data-transfers).

roeierez: I don't understand why we need to use work manager here: https://github.com/breez/c-breez/blob/main/android/app/src/main/java/com/cBreez/client/JobManager.kt#L60

If we are using a service shouldn't we just run the code as we already get the cpu and time we need?

roeierez: > I suppose type of our foreground service would be considered [`dataSync`](https://developer.android.com/about/versions/14/changes/fgs-types-required#data-sync), which will be also be deprecated and WorkManager is the most fitting of the suggested alternatives.
> 
> We could decide whether if there's a limiting factor in using [`shortService`](https://developer.android.com/about/versions/14/changes/fgs-types-required#short-service) type for our job and forgo WorkManager.

I don't think dataSync is the right type as dataSync is something that happens periodically and not critical. It seems that shortService is more suitable.

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
```diff
@@ -0,0 +1,240 @@
...
+        return START_NOT_STICKY
+    }
+
+    private fun startBusiness(
+        paymentHash: String,
+        clickAction: String?,
+    ) {
+        val notification = notifyForegroundService(applicationContext)
+        when {
+            breezSDK != null && _state.value is NodeServiceState.Running -> {
```
roeierez, line 158: This whole state management seems complicated. Why do we need it (NodeState)?
What's wrong in simple callling connectSDK (which already handles a singleton static instance) ?

roeierez (changes requested):

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
```diff
@@ -0,0 +1,196 @@
...
+            registerNotificationChannels(applicationContext)
+        }
+        Logger.tag(TAG).debug { "Breez node service created." }
+    }
+
+    // =========================================================== //
+    //                      SERVICE LIFECYCLE                      //
+    // =========================================================== //
+
+    override fun onBind(intent: Intent?): IBinder {
```
roeierez, line 73: Can you explain why we stop the foreground service on bind ?

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
```diff
@@ -0,0 +1,196 @@
...
+                return
+            } else {
+                Logger.tag(TAG).info { "Payment w/ paymentHash: $paymentHash not received yet." }
+                withContext(Dispatchers.IO) { Thread.sleep(1_000) }
+            }
+        }
+
+        // If we reach here then we didn't receive for more than 1 minute
+        paymentHashPollerList.remove(paymentHash)
+        throw Exception("Payment not found before timeout")
```
roeierez, line 170: Where is this exception handled?

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
```diff
@@ -0,0 +1,196 @@
...
+    // =========================================================== //
+
+    /** Called when an intent is called for this service. */
+    @RequiresApi(Build.VERSION_CODES.TIRAMISU)
+    override fun onStartCommand(intent: Intent?, flags: Int, startId: Int): Int {
+        super.onStartCommand(intent, flags, startId)
+        Logger.tag(TAG).debug {
+            "Start Breez node service from intent [ intent=$intent, flag=$flags, startId=$startId ]"
+        }
+        intent.getRemoteMessage()?.let {
```
roeierez, line 113: Seems like we should schedule a shutdown here otherwise there are cases where the service will not stop.

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
```diff
@@ -0,0 +1,196 @@
...
+                                    Logger.tag(TAG).error { "Error when polling payment: $e" }
+                                    notifyPaymentFailed(applicationContext)
+                                }
+                    ) {
+                        pollPaymentHash(paymentHash, clickAction)
+                    }
+                }
+            }
+        }
+        return START_NOT_STICKY
```
roeierez, line 135: Seems like we should schedule a shutdown here otherwise there are cases where the service will not stop.

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
```diff
@@ -0,0 +1,196 @@
...
+    ) {
+        Logger.tag(TAG).info {
+            "Received payment. Bolt11:${bolt11}\nPayment Hash:${paymentHash}"
+        }
+        val amountSat = (amountMsat ?: ULong.MIN_VALUE) / 1000u
+        notifyPaymentReceived(applicationContext, clickAction = clickAction, amountSat = amountSat)
+
+        // Push back shutdown by SHUTDOWN_DELAY_MS in case we'll receive more payments
+        shutdownHandler.removeCallbacksAndMessages(null)
+        shutdownHandler.postDelayed(shutdownRunnable, SHUTDOWN_DELAY_MS)
```
roeierez, line 187: Seems like we will not get any "Failed payment" notification but I think it is good behavior that also simplify the flow so let's keep it this way.

android/app/src/main/java/com/cBreez/client/DismissNotification.kt
```diff
@@ -0,0 +1,13 @@
+package com.cBreez.client
+
+import android.content.BroadcastReceiver
+import android.content.Context
+import android.content.Intent
+import androidx.core.app.NotificationManagerCompat
+import com.cBreez.client.Constants.NOTIFICATION_ID_PAYMENT_FAILED
+
+class DismissNotification : BroadcastReceiver() {
```
roeierez, line 9: Do we use this or can be removed?

android/app/src/main/kotlin/com/cBreez/client/MainActivity.kt
```diff
@@ -1,5 +1,52 @@
...
+import org.tinylog.kotlin.Logger
 
-class MainActivity : FlutterActivity()
\ No newline at end of file
+class MainActivity : FlutterActivity() {
+    private val _service = MutableLiveData<BreezForegroundService?>(null)
+
+    override fun onStart() {
+        super.onStart()
+        Intent(this, BreezForegroundService::class.java).let { intent ->
```
roeierez, line 22: Why do we need to bind and unbind to the service from the app?

roeierez (changes requested):

android/app/src/main/java/com/cBreez/client/DismissNotification.kt
erdemyerebasmaz: Removed as part of [Remove payment hash poller & everything related to failed payment not…](https://github.com/breez/c-breez/pull/748/commits/fc1bb41f68315a4d63fbac3955c802d002d00a73)

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
erdemyerebasmaz: Addressed as part of [Remove payment hash poller & everything related to failed payment not…](https://github.com/breez/c-breez/pull/748/commits/fc1bb41f68315a4d63fbac3955c802d002d00a73)

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
erdemyerebasmaz: https://github.com/breez/c-breez/pull/748#discussion_r1439530834

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
erdemyerebasmaz: Saw these comments a bit late after making changes... 

We're no longer polling for payment hash & displaying any failed payment notifications anymore.

We're delaying the shutdown by 2 minutes every time we receive a payment but this was taking poller into consideration, which ran for a 1 minute, so we should reduce shutdown delay back to 1 minute.

roeierez:
Looks good. I will test it and report back.

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
```diff
@@ -0,0 +1,148 @@
...
+    inner class SDKListener : EventListener {
+        override fun onEvent(e: BreezEvent) {
+            Logger.tag(TAG).info { "Received event $e" }
+            when (e) {
+                is BreezEvent.InvoicePaid -> {
+                    val pd = e.details
+                    handleReceivedPayment(pd.bolt11, pd.paymentHash, pd.payment?.amountMsat)
+
+                    // Push back shutdown by SHUTDOWN_DELAY_MS in case we'll receive more payments
+                    pushbackShutdown()
```
roeierez, line 42: I don't think we need that. It makes the foreground service keep running in the task bar.
Also we need to wait for sync after InvoicePaid, see the corresponding implementation in iOS.

roeierez (changes requested):

erdemyerebasmaz:
> ### TODO: 
> - [ ] Enable autostart &
> - [ ] Disable battery optimizations
> 
> for the app.

These items will be handled in another PR, related branch: [android_notifications_autostart_battery_optimizations](https://github.com/breez/c-breez/tree/android_notifications_autostart_battery_optimizations)

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
```diff
@@ -0,0 +1,167 @@
...
+        connectSdkIfNeeded(intent)
+
+        // Push back shutdown by SHUTDOWN_DELAY_MS after connecting to the SDK
+        pushbackShutdown()
+        return START_NOT_STICKY
+    }
+
+    /** Checks whether if the data payload is of type payment_received & has a payment hash */
+    private fun connectSdkIfNeeded(intent: Intent?) {
+        intent?.remoteMessage?.run {
```
roeierez, line 120: It seems cleaner to check this on the higher level (onStartCommand) and then shutdown in case the message is not recognized.
Right now the process keeps running until the delay ends.

roeierez (changes requested):

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
erdemyerebasmaz: Done 79d7566aebf9ea00d6c65143b6a019fbb72f7a5c

android/app/src/main/java/com/cBreez/client/BreezForegroundService.kt
erdemyerebasmaz: Done [Shutdown service after receiving synced event for received payment](https://github.com/breez/c-breez/pull/748/commits/5feeb18c39e1cd8f48d75183afeff74706e00373)

made a little change later to only log if payment is received https://github.com/breez/c-breez/pull/748/commits/79d7566aebf9ea00d6c65143b6a019fbb72f7a5c

android/app/src/main/AndroidManifest.xml
```diff
@@ -165,14 +173,6 @@
...
             android:name="com.google.firebase.messaging.default_notification_color"
             android:resource="@color/breez_notification_color" />
-        <!-- Starting in Android 8.0 (API level 26), all notifications must be assigned to a channel
-             or they don't appear. This lets users disable specific notification channels for your app
-             instead of disabling all your notifications. Users can control the visual and auditory
-             options for each channel from the Android system settings, as shown in figure 11.
-             Users can also touch & hold a notification to change behaviors for the associated channel. -->
-        <meta-data
-            android:name="com.google.firebase.messaging.default_notification_channel_id"
-            android:value="@string/offline_payments_notification_channel_id" />
```
erdemyerebasmaz, line 175: We're only interested in showing foreground service status & payment received notifications and we're registering only to their notification channels. This part is no longer needed.

roeierez (approved):
LGTM

erdemyerebasmaz:
> > ### TODO: 
> 
> > - [ ] Enable autostart &
> 
> > - [ ] Disable battery optimizations
> 
> > 
> 
> > for the app.
> 
> 
> 
> These items will be handled in another PR, related branch: [android_notifications_autostart_battery_optimizations](https://github.com/breez/c-breez/tree/android_notifications_autostart_battery_optimizations)

Other items on to-do list:

- Set SDK log stream
- Disconnect from SDK on shutdown

PR #750, opened 2023-12-28
Add pendingExpirationTimestamp information

ubbabeck:
# About pr

Implements  and depends on https://github.com/breez/breez-sdk/pull/674

- Adds `pendingExpirationBlock` to the PaymentMinutiae.

- Creates function to convert block difference to time.

I decided to have to pendingExpirationBlock in the payment field and not the calculated approximate timestamp as we might want to use the `pendingExpirationBlock` for something else at a later stage. Let me know if we'd rather want to add `pendingExpirationTime` to PaymentMinutiae instead.

### DIFF
--- a/lib/routes/home/widgets/payments_list/dialog/payment_details_dialog_expiration.dart
+++ b/lib/routes/home/widgets/payments_list/dialog/payment_details_dialog_expiration.dart
@@ -1,9 +1,11 @@
 import 'package:auto_size_text/auto_size_text.dart';
 import 'package:breez_sdk/bridge_generated.dart';
 import 'package:breez_translations/breez_translations_locales.dart';
+import 'package:c_breez/bloc/account/account_bloc.dart';
 import 'package:c_breez/models/payment_minutiae.dart';
 import 'package:c_breez/utils/date.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter_bloc/flutter_bloc.dart';
 
 class PaymentDetailsDialogExpiration extends StatelessWidget {
   final PaymentMinutiae paymentMinutiae;
@@ -24,6 +26,9 @@ class PaymentDetailsDialogExpiration extends StatelessWidget {
       return Container();
     }
 
+    final pendingExpirationBlock = paymentMinutiae.pendingExpirationBlock;
+    final currentBlockHeight = context.select<AccountBloc, int>((value) => value.state.blockheight);
+
     return Container(
       height: 36.0,
       padding: const EdgeInsets.only(left: 16.0, right: 16.0, top: 8.0),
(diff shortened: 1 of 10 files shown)


lib/routes/home/widgets/payments_list/dialog/payment_details_dialog_expiration.dart
```diff
@@ -39,7 +44,21 @@ class PaymentDetailsDialogExpiration extends StatelessWidget {
...
+                pendingExpirationTime != null
+                    ? BreezDateUtils.formatYearMonthDayHourMinute(pendingExpirationTime)
+                    : "",
+                style: themeData.primaryTextTheme.displaySmall,
+                textAlign: TextAlign.right,
+                maxLines: 1,
+              ),
+            ),
+          )
         ],
```
erdemyerebasmaz, line 62: Maybe it's better to use a helper method for pending expiration time so the logic is more compact.

```
AutoSizeText(
    _getExpirationTime,
    ...
...
_getExpirationTime() {
    final blockHeight = context.read<AccountBloc>().state.blockheight;
    final pendingExpirationBlock = paymentMinutiae.pendingExpirationBlock;
    if(pendingExpirationBlock != null) {
        final pendingExpirationTime = BreezDateUtils.blockDiffToDate(
            blockHeight: blockHeight,
            expiryBlock: pendingExpirationBlock,
        );
        return BreezDateUtils.formatYearMonthDayHourMinute(pendingExpirationTime);
    }
   return "";
}
```

lib/utils/date.dart
```diff
@@ -39,6 +39,18 @@ class BreezDateUtils {
...
+      int diff = expiryBlock - blockHeight;
+      if (diff > 0) {
+        diff *= 600;
+        final time = DateTime.now();
+        return time.add(Duration(seconds: diff));
+      }
+    }
+    return null;
+  }
+
```
erdemyerebasmaz, line 53: Let's make this non-nullable.
```suggestion
  static DateTime blockDiffToDate({required int blockHeight, required int expiryBlock}) {
      int diff = expiryBlock - blockHeight;
      if (diff > 0) {
        diff *= 600;
        final time = DateTime.now();
        return time.add(Duration(seconds: diff));
      }
  }

```

erdemyerebasmaz (commented):
Looks good
Have 2 nit comments

lib/utils/date.dart
erdemyerebasmaz: I suppose it should return current date if there's no difference.

lib/routes/home/widgets/payments_list/dialog/payment_details_dialog_expiration.dart
ubbabeck: Removed this logic and added a `pendingExpirationTime` to the PaymentMinutiae also.

lib/utils/date.dart
```diff
@@ -40,15 +40,11 @@ class BreezDateUtils {
...
-        final time = DateTime.now();
-        return time.add(Duration(seconds: diff));
-      }
-    }
-    return null;
+    if (expiryBlock == null || expiryBlock <= blockHeight) return null;
+    int diff = expiryBlock - blockHeight;
+    diff *= 600;
+    final time = DateTime.now();
+    return time.add(Duration(seconds: diff));
```
erdemyerebasmaz, line 47: Avoid redundant calculation for `diff` variable.
```suggestion
    int diffInSecs = (expiryBlock - blockHeight) * 600;
    final time = DateTime.now();
    return time.add(Duration(seconds: diffInSecs));
```

lib/routes/home/widgets/payments_list/payment_item_amount.dart
```diff
@@ -97,105 +97,109 @@ void main() async {
...
+                        keysend: false,
+                        bolt11: "",
+                      ),
                     ),
                   ),
-                ),
-                getSystemAppLocalizations(),
-              ),
+                  getSystemAppLocalizations(),
+                  800000),
```
erdemyerebasmaz, line 148: Can you add a comma after these values and reformat code? It helps with readability & keeping the right indentation.

lib/utils/date.dart
ubbabeck: Done

lib/routes/home/widgets/payments_list/payment_item_amount.dart
ubbabeck: Added

lib/utils/date.dart
ubbabeck: Added

ademar111190 (approved):

roeierez:
@ubbabeck @erdemyerebasmaz let's move this forward, seems like not much (if any )work left.

erdemyerebasmaz (approved):
LGTM
Please address CI issue and the typo before merging.

PR #751, opened 2023-12-28
sync after offline payment receive

### LINKED ISSUE #736: Balance update
When Cbreez receives a payment while it is in the background the balance does not update when the app is opened.

ubbabeck:
fixes #736 for iOS

Sync after payment is received in order to let the sdk cache the latest nodestate.

### DIFF
--- a/ios/Breez Notification Service Extension/NotificationService.swift
+++ b/ios/Breez Notification Service Extension/NotificationService.swift
@@ -106,6 +106,7 @@ class NotificationService: UNNotificationServiceExtension {
             }
         }
         self.paymentReceivers.removeAll(where: {$0.paymentHash == data.paymentHash})
+        try? NotificationService.breezSDK!.sync()
     }
 }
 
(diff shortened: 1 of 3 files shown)


erdemyerebasmaz (approved):
LGTM!

ios/Breez Notification Service Extension/NotificationService.swift
```diff
@@ -106,6 +106,7 @@ class NotificationService: UNNotificationServiceExtension {
             }
         }
         self.paymentReceivers.removeAll(where: {$0.paymentHash == data.paymentHash})
+        try? NotificationService.breezSDK!.sync()
```
roeierez, line 109: This will trigger another sync that is already in progress. I think we should wait for that sync to complete instead of triggering another one

roeierez (changes requested):

ios/Breez Notification Service Extension/NotificationService.swift
ubbabeck: That sounds like a better solution, will add a waiting mechanism for a sync. 

@roeierez  So we will have to wait for a sync event after payment succeed and not trigger on the first event which is triggered by connect?

roeierez: Yes, seems like this is the correct path.

ubbabeck:
@roeierez are refactoring offline receive for iOS so he will also handle sync after payment receive. Closing this pr.

PR #752, opened 2024-01-02
Single extension instance

roeierez:
This PR simplifies the notification extension logic and prepares it to other upcoming types of background tasks (such as lnurlpay)
For each notification type a proper task is created and being started to handle that notification end to end.
Also found out that the notification is not invoked in parallel. If there are multiple payments in parallel, the notifications are queued and invoked one after another so no need to code for the case where we can several notifications for the same instance.
I also noticed we don't need the payment hash poller, we can just wait for payment received event which seems to be reliable enough.
fixed #751

### DIFF
--- a/ios/Breez Notification Service Extension/BreezManager.swift
+++ b/ios/Breez Notification Service Extension/BreezManager.swift
@@ -11,30 +11,7 @@ fileprivate var log = Logger(
 fileprivate var log = Logger(OSLog.disabled)
 #endif
 
-// SDK events listener
-class SDKListener: EventListener {
-    
-    var paymentListener : PaymentListener
-    
-    init(paymentListener: @escaping PaymentListener) {
-        self.paymentListener = paymentListener
-    }
-    
-    func onEvent(e: BreezEvent) {
-        switch e {
-        case .invoicePaid(details: let details):
-            log.info("Received payment. Bolt11: \(details.bolt11)\nPayment Hash:\(details.paymentHash)")
-            if details.payment != nil {
-                paymentListener(details.payment!)
-            }
-            return
-        default:
-            break
-        }
-    }
-}
-
-func connectSDK(paymentListener: @escaping PaymentListener) throws -> BlockingBreezServices? {
+func connectSDK(eventListener: EventListener) throws -> BlockingBreezServices? {
     log.trace("connectSDK()")
     
     // Create the default config
@@ -57,7 +34,7 @@ func connectSDK(paymentListener: @escaping PaymentListener) throws -> BlockingBr
         throw SdkError.Generic(message: "seed not found")
     }
     log.trace("Connecting to Breez SDK")
-    let breezSDK = try connect(config: config, seed: seed!, listener: SDKListener(paymentListener: paymentListener))
+    let breezSDK = try connect(config: config, seed: seed!, listener: eventListener)
     log.trace("Connected to Breez SDK")
     return breezSDK
 }
--- a/ios/Breez Notification Service Extension/BreezManager.swift
+++ b/ios/Breez Notification Service Extension/BreezManager.swift
@@ -11,32 +11,66 @@ fileprivate var log = Logger(
 fileprivate var log = Logger(OSLog.disabled)
 #endif
 
-func connectSDK(eventListener: EventListener) throws -> BlockingBreezServices? {
-    log.trace("connectSDK()")
+class BreezManager {
+    private static var breezSDK: BlockingBreezServices? = nil
+    fileprivate static var queue = DispatchQueue(label: "BreezManager")
+    fileprivate static var sdkListener: EventListener? = nil
     
-    // Create the default config
-    let apiKey = try Environment.glApiKey()
-    log.trace("API_KEY: .\(apiKey)")
-    var config = defaultConfig(envType: EnvironmentType.production, apiKey: apiKey,
-                               nodeConfig: NodeConfig.greenlight(
-                                config: GreenlightNodeConfig(partnerCredentials: nil, inviteCode: nil)))
-    
-    config.workingDir = FileManager
-        .default.containerURL(forSecurityApplicationGroupIdentifier: "group.F7R2LZH3W5.com.cBreez.client")!
-        .absoluteString
-    
-    // Construct the seed
-    let mnemonic = CredentialsManager.shared.restoreMnemonic() ?? ""
-    log.trace("mnemonic: .\(mnemonic)")
-    let seed = try? mnemonicToSeed(phrase: mnemonic)
-    // Connect to the Breez SDK make it ready for use
-    guard seed != nil else {
-        throw SdkError.Generic(message: "seed not found")
+    static func register(listener: EventListener) throws -> BlockingBreezServices? {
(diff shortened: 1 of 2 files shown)


ios/Breez Notification Service Extension/BreezManager.swift
```diff
@@ -11,55 +11,66 @@ fileprivate var log = Logger(
...
+    fileprivate static var sdkListener: EventListener? = nil
     
-    var paymentListener : PaymentListener
+    static func register(listener: EventListener) throws -> BlockingBreezServices? {
+        try BreezManager.queue.sync { [] in
+            BreezManager.sdkListener = listener
+            if BreezManager.breezSDK == nil {
+                BreezManager.breezSDK = try BreezManager.connectSDK()
+            }
+            return BreezManager.breezSDK
```
erdemyerebasmaz, line 25:
```suggestion
            return BreezManager.breezSDK ?? (BreezManager.breezSDK = try BreezManager.connectSDK())
```

ios/Breez Notification Service Extension/BreezManager.swift
```diff
@@ -11,55 +11,66 @@ fileprivate var log = Logger(
...
+        var config = defaultConfig(envType: EnvironmentType.production, apiKey: apiKey,
+                                   nodeConfig: NodeConfig.greenlight(
+                                    config: GreenlightNodeConfig(partnerCredentials: nil, inviteCode: nil)))
+        
+        config.workingDir = FileManager
+            .default.containerURL(forSecurityApplicationGroupIdentifier: "group.F7R2LZH3W5.com.cBreez.client")!
+            .absoluteString
+        
+        // Construct the seed
+        let mnemonic = CredentialsManager.shared.restoreMnemonic() ?? ""
```
erdemyerebasmaz, line 51: If mnemonic is empty, it will fail to generate a seed, let's convert this into a guard statement
```suggestion
        guard let mnemonic = CredentialsManager.shared.restoreMnemonic() else {
            throw SdkError.Generic(message: "mnemonic not found")
        }
```

ios/Breez Notification Service Extension/BreezManager.swift
```diff
@@ -11,55 +11,66 @@ fileprivate var log = Logger(
...
+        log.trace("mnemonic: .\(mnemonic)")
+        let seed = try? mnemonicToSeed(phrase: mnemonic)
+        // Connect to the Breez SDK make it ready for use
+        guard seed != nil else {
+            throw SdkError.Generic(message: "seed not found")
         }
```
erdemyerebasmaz, line 57: This can be reduced into a single guard statement and in light of converting mnemonic into a guard statement, we can use a better representative error message here.
```suggestion
        guard let seed = try? mnemonicToSeed(phrase: mnemonic) else {
            throw SdkError.Generic(message: "failed to generate seed")
        }
        // Connect to the Breez SDK make it ready for use
```
and the comment line has to be moved below.

ios/Breez Notification Service Extension/BreezManager.swift
```diff
@@ -11,55 +11,66 @@ fileprivate var log = Logger(
...
-    let apiKey = try Environment.glApiKey()
-    log.trace("API_KEY: .\(apiKey)")
-    var config = defaultConfig(envType: EnvironmentType.production, apiKey: apiKey,
-                               nodeConfig: NodeConfig.greenlight(
-                                config: GreenlightNodeConfig(partnerCredentials: nil, inviteCode: nil)))
+    func onEvent(e: BreezEvent) {
+        BreezManager.queue.async { [] in
+            if let l = BreezManager.sdkListener {
+                l.onEvent(e: e)
+            }
```
erdemyerebasmaz, line 71: Can we use optional chaining here?
```suggestion
            BreezManager.sdkListener?.onEvent(e: e)
```

ios/Breez Notification Service Extension/NotificationService.swift
```diff
@@ -87,41 +75,51 @@ class NotificationService: UNNotificationServiceExtension {
...
-        self.paymentHash = paymentHash
         self.logger = logger
     }
     
+    func onShutdown() {
+        if let p =  self.receivedPayment {
+            self.displayPushNotification(title: "Received \(p.amountMsat/1000) sats")
+        } else {
+            self.displayPushNotification(title: "Receive payment failed")
+        }
```
erdemyerebasmaz, line 102: I don't remember the exact syntax but could be reduced to sth like
```suggestion
        let title = self.receivedPayment.first { "Received \($0.amountMsat/1000) sats" } ??  "Receive payment failed"
        self.displayPushNotification(title: title)
```

erdemyerebasmaz (commented):
Looks good, have some syntactic nit comments

ios/Breez Notification Service Extension/BreezManager.swift
roeierez: Doesn't compile for me

ios/Breez Notification Service Extension/NotificationService.swift
roeierez: receivedPayment is not an array so no first so I changes it slightly different.

ios/Breez Notification Service Extension/BreezManager.swift
roeierez: Done

ios/Breez Notification Service Extension/BreezManager.swift
roeierez: Done

ios/Breez Notification Service Extension/BreezManager.swift
roeierez: Done

ios/Breez Notification Service Extension/NotificationService.swift
```diff
@@ -15,70 +19,54 @@ class NotificationService: UNNotificationServiceExtension {
...
-                            self?.onPaymentReceived(payment: payment)
-                        }
-                    })
-                    self.logger.info("Breez SDK connected successfully")
+                    Self.logger.info("Breez SDK is not connected, connecting....")
+                    self.breezSDK = try BreezManager.register(listener: currentTask)
+                    Self.logger.info("Breez SDK connected successfully")
                 } catch {
-                    self.logger.info("Breez SDK connections failed \(error)")
+                    Self.logger.info("Breez SDK connections failed \(error)")
```
erdemyerebasmaz, line 45: I'm trying to match Android service logs with iOS extension and noticed a typo & a wrong log level here

```suggestion
                    Self.logger.error("Breez SDK connection failed \(error)")
```

ios/Breez Notification Service Extension/BreezManager.swift
erdemyerebasmaz: Right, don't think we can use `try` on right side of a nil coalescing operator... Ignore this.

ios/Breez Notification Service Extension/NotificationService.swift
roeierez: Done

erdemyerebasmaz (approved):

Issue #729, opened 2023-12-01
Implement Health Check Status 


ubbabeck:
ref https://github.com/breez/breez-sdk/pull/628

PR #758, opened 2024-01-09
Implement Health Check Status

ademar111190:
Implement Health Check Status
Fixes https://github.com/breez/c-breez/issues/729

# How it looks like

|Maintanance|Service disruption|
|:-:|:-:|
|maintanance|disruption|

# How to test

## Method 1 (changing the code)
- Change the code to emit a different `HealthCheckStatus` than `Operational`
- You should see at the home page a flushbar like that one in screenshot

## Method 2 (being patient)
- Wait for a moment in the future when breez service is in maintenance or has it service disrupted
- You should see at the home page a flushbar like that one in screenshot

### DIFF
--- a/lib/handlers/health_check_handler.dart
+++ b/lib/handlers/health_check_handler.dart
@@ -11,6 +11,7 @@ import 'package:flutter/material.dart';
 import 'package:flutter_bloc/flutter_bloc.dart';
 import 'package:flutter_svg/flutter_svg.dart';
 import 'package:logging/logging.dart';
+import 'package:rxdart/rxdart.dart';
 
 final _log = Logger("HealthCheckHandler");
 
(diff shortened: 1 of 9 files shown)


erdemyerebasmaz:
@roeierez  @dangeross Is it possible to convert this function into a stream? It seems more intuitive than setting up a polling mechanism.

lib/handlers/health_check_handler.dart
```diff
@@ -0,0 +1,66 @@
...
+      return;
+    }
+
+    final texts = context.texts();
+    final themeData = Theme.of(context);
+    final bloc = context.read<HealthCheckBloc>();
+
+    _listener?.cancel();
+    _listener = bloc.stream.listen((event) {
+      if (event == HealthCheckStatus.Maintenance || event == HealthCheckStatus.ServiceDisruption) {
```
roeierez, line 35: Shouldn't we also remove the flushbar when the helth check is normal again?

roeierez:
> @roeierez @dangeross Is it possible to convert `serviceHealthCheck` function into a stream? It seems more intuitive than setting up a polling mechanism.

It means the SDK either needs to poll or maintain a live connection to the breez server. I think both are overkill.
IMO we don't need to poll here wither, certainly not every 30 seconds. Perhaps only check on restart the app? What are we doing in breez mobile?
@ademar111190 what do you think about adding a button on the flashbar to re-check the status?

roeierez (changes requested):

ademar111190:
@roeierez @erdemyerebasmaz  breez mobile pulls every 30 seconds; I copied the logic from there. Honestly, I believe that checking it one time once the app opens is enough, but maybe I am missing something that was discussed at the time people built it on breez mobile.

Should I change it to check at the time the app opens only? If yes, should I change it for breez mobile too?

lib/handlers/health_check_handler.dart
ademar111190: good catch; fixed.

dangeross:
Just as a thought - maybe when a different request to the SDK fails, we can poll the health status?

roeierez:
> Just as a thought - maybe when a different request to the SDK fails, we can poll the health status? 

In theory you are right @dangeross but in practice I think there are many edge cases such as API to the lsp and server works perfectly but the lsp node is down for maintenance so I suspect it doesn't worth the effort.

roeierez:
> @roeierez @erdemyerebasmaz  breez mobile pulls every 30 seconds; I copied the logic from there. Honestly, I believe that checking it one time once the app opens is enough, but maybe I am missing something that was discussed at the time people built it on breez mobile.
> 
> Should I change it to check at the time the app opens only? If yes, should I change it for breez mobile too?

No need to change breez mobile.
How about one time check on start and poll every minute if the status is not normal? Once normal remove the flushbar.
Also IMO good to have a button in the flashbar to refresh the status manually.

ademar111190:
Ok; not changing breez mobile.

I did apply your suggestions @roeierez also I moved the call trigger to a separate method; so if we want to trigger it again after some fail or other flow as @dangeross has suggested, we just need to get the bloc and trigger the call.

roeierez (approved):
LGTM

PR #759, opened 2024-01-10
Lnurl pay

roeierez:


lib/bloc/webhooks/webhooks_bloc.dart
```diff
@@ -0,0 +1,185 @@
...
+        lnurlPayError: lnurlPayError ?? this.lnurlPayError,
+        loading: loading ?? this.loading);
+  }
+}
+
+class WebhooksBloc extends Cubit<WebhooksState> {
+  static const notifierServiceURL = "https://notifier.breez.technology";
+  static const lnurlServiceURL = "https://lnurl.breez.technology";
+
+  final _log = Logger("LnurlPayBloc");
```
erdemyerebasmaz, line 59:
```suggestion
  final _log = Logger("WebhooksBloc");
```

lib/bloc/webhooks/webhooks_bloc.dart
```diff
@@ -0,0 +1,185 @@
...
+        : defaultTargetPlatform == TargetPlatform.android
+            ? "android"
+            : "";
+    if (platform.isEmpty) {
+      throw Exception("Notifications for platform is not supported");
+    }
+    return "$notifierServiceURL/api/v1/notify?platform=$platform&token=$token";
+  }
+
+  String generateUser() {
```
erdemyerebasmaz, line 159: This function is never used and it's better off being an utility function under utils folder.

lib/routes/lightning_address/lightning_address_page.dart
```diff
@@ -0,0 +1,79 @@
...
+  }
+}
+
+class LightningAddressPageState extends State<LightningAddressPage> {
+  @override
+  void initState() {
+    super.initState();
+    final webhookBloc = context.read<WebhooksBloc>();
+    webhookBloc.refreshLnurlPay();
+    context.read<InputBloc>().trackPayment(null).then((value) {
```
erdemyerebasmaz, line 31: What if we receive another payment? How does this work in terms of reliability?

erdemyerebasmaz (changes requested):
Looks good overall, have some nit comments

lib/bloc/webhooks/webhooks_bloc.dart
roeierez: fixed

lib/bloc/webhooks/webhooks_bloc.dart
```diff
@@ -0,0 +1,185 @@
...
+  }
+}
+
+class WebhooksBloc extends Cubit<WebhooksState> {
+  static const notifierServiceURL = "https://notifier.breez.technology";
+  static const lnurlServiceURL = "https://lnurl.breez.technology";
+
+  final _log = Logger("LnurlPayBloc");
+  final BreezSDK _breezSDK;
+  final ServiceInjector injector;
```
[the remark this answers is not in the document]
roeierez: fixed

lib/bloc/webhooks/webhooks_bloc.dart
roeierez: Right, sneaked by mistake.

lib/routes/lightning_address/lightning_address_page.dart
roeierez: We don't really have a way to associate a specific payment. Worst case (which is very rare) we will exit this screen on payment and show the payment receive screen (which sounds OK to me).

erdemyerebasmaz: Perhaps we can pass a flag here, `isLnurlPay` and check if the next received payment is a ln payment on trackPayment as we have payment details there. Either way, it's OK for me.

erdemyerebasmaz (approved):
LGTM

Have one optional request: https://github.com/breez/c-breez/pull/759#discussion_r1463356717

lib/routes/lightning_address/lightning_address_page.dart
roeierez: The trackPayment always notifies only for Lnpayment so we don't need the flag.

PR #765, opened 2024-01-23
Wait for log file initialization before getting build and device info

ademar111190:
minor improvements on log flow

- Change the file name of the log file for something more human-readable
- Wait for log file initialization before:
  - registering the error listener
  - getting the git info
  - getting the device info


lib/logger.dart
```diff
@@ -42,7 +42,7 @@ class BreezLogger {
     Config.instance().then((config) {
       var appDir = Directory(config.sdkConfig.workingDir);
       _pruneLogs(appDir);
-      final file = File("${_logDir(appDir)}/${DateTime.now().millisecondsSinceEpoch}.app.log");
+      final file = File("${_logDir(appDir)}/${_formatTime(DateTime.now())}.log");
```
erdemyerebasmaz, line 45: keep the `.app.` part

ademar111190: Ok; but for curiosity, why?

erdemyerebasmaz: It's an identifier we've added to distinguish app logs from extension logs instead of using a separate folder for them.

erdemyerebasmaz (approved):
LGTM

PR #772, opened 2024-01-29
Cleanup on the home drawer menu

ademar111190:
A minor cleanup on the home drawer menu:

- Remove unused code inherited from Breez mobile
- Remove "builders" method in favor of creating widgets in the build method directly
- Break some widgets into their own files to reduce cognitive loading when opening a file


roeierez:
@ademar111190  breaking down to different widgets makes sense to me. Using too many files sometimes makes it harder to maintain the code as we need to switch back and forth different files in order to understand the code.
I don't have a clear opinion but I think I am in favor of trying to find a way to use less files by grouping some widgets that are built together into one file.

PR #774, opened 2024-01-30
add pending balance field

ubbabeck:
*adds new pending balance field to node state.
*update dependencies


ademar111190 (approved):
Thank you 🙏

roeierez (approved):

PR #781, opened 2024-02-14
Allow setting optional checkout reference for SDK on CI workflows

erdemyerebasmaz:
This PR addresses #780

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -11,6 +11,7 @@ on:
         description: 'SDK commit/tag/branch reference'
         required: false
         type: string
+        default: 'main'
 
 jobs:
   build:
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,22 +1,20 @@
 name: Run CI
-
-# Controls when the workflow will run
 on:
   # Triggers the workflow on push events but only for the "main" branch
   push:
     branches: [ "main" ]
   pull_request:
     branches: [ "main" ]
-
-  # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
+    inputs:
+      ref:
+        description: 'SDK commit/tag/branch reference'
+        required: false
+        type: string
 
 jobs:
-
   build:
-
     runs-on: macOS-13
-
     steps:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


dangeross (approved):
LGTM. I'm assuming an empty ref in the checkout defaults to main

.github/workflows/CI.yml
```diff
@@ -28,6 +26,7 @@ jobs:
           repository: 'breez/breez-sdk'
           ssh-key: ${{secrets.REPO_SSH_KEY}}
           path: 'breez-sdk'
+          ref: ${{ inputs.ref }}
```
JssDWt, line 29: There won't be an input on `push` or `pull request`, so perhaps add a fallback to the github sha to make sure the field is always set.

```suggestion
          ref: ${{ inputs.ref || github.sha }}
```

.github/workflows/build-android.yml
```diff
@@ -57,6 +53,7 @@ jobs:
           repository: 'breez/breez-sdk'
           ssh-key: ${{secrets.REPO_SSH_KEY}}
           path: 'breez-sdk'
+          ref: ${{ inputs.ref }}
```
JssDWt, line 56: I would suggest the same here

```suggestion
          ref: ${{ inputs.ref || github.sha }}
```

.github/workflows/build-ios.yml
```diff
@@ -75,6 +71,7 @@ jobs:
           repository: 'breez/breez-sdk'
           ssh-key: ${{secrets.REPO_SSH_KEY}}
           path: 'breez-sdk'
+          ref: ${{ inputs.ref }}
```
JssDWt, line 74:
```suggestion
          ref: ${{ inputs.ref || github.sha }}
```

JssDWt (commented):
some comments about fallback if the CI isn't run through workflow_dispatch

erdemyerebasmaz:
> LGTM. I'm assuming an empty ref in the checkout defaults to main

@dangeross It is according to docs
```
# The branch, tag or SHA to checkout. When checking out the repository that
# triggered a workflow, this defaults to the reference or SHA for that event.
# Otherwise, uses the default branch.
ref: ''
```

Since we're checking out a different repository, not the one that triggered the workflow, it should use the default branch.

JssDWt:
> LGTM. I'm assuming an empty ref in the checkout defaults to main

Maybe we can make this explicit, just to be sure?

```
${{ inputs.ref || 'main' }}
```

erdemyerebasmaz:
> > LGTM. I'm assuming an empty ref in the checkout defaults to main
> 
> Maybe we can make this explicit, just to be sure?
> 
> ```
> ${{ inputs.ref || 'main' }}
> ```

Sure, I'll explicitly set the default value to `main`

PR #783, opened 2024-02-15
Handle notifications for swap txs

### LINKED ISSUE #778: Handle swap in notifications
We need to match the behavior to breez mobile.
Once a swap notification is received we need to:
1. Explicitly try to redeem the swap funds (complete swap)
2. Open the app.

ok300:
Fixes #778

### DIFF
--- a/android/app/src/main/kotlin/com/cBreez/client/job/RedeemSwap.kt
+++ b/android/app/src/main/kotlin/com/cBreez/client/job/RedeemSwap.kt
@@ -1,0 +1,55 @@
+package com.cBreez.client.job
+
+import android.content.Context
+import breez_sdk.BlockingBreezServices
+import breez_sdk.BreezEvent
+import breez_sdk.SdkException
+import com.cBreez.client.BreezNotificationHelper.Companion.notifyChannel
+import com.cBreez.client.Constants
+import com.cBreez.client.ForegroundService
+import com.cBreez.client.R
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.encodeToString
+import kotlinx.serialization.json.Json
+import org.tinylog.kotlin.Logger
+
+@Serializable
+data class AddressTxsConfirmedRequest(
+    @SerialName("address") val address: String,
+)
+
+class RedeemSwapJob(
+    private val context: Context,
+    private val fgService: ForegroundService,
+    private val payload: String,
+) : SDKJob {
+    companion object {
+        private const val TAG = "RedeemSwapJob"
+    }
+
+    override fun start(breezSDK: BlockingBreezServices) {
+        val request = Json.decodeFromString(AddressTxsConfirmedRequest.serializer(), payload)
+
+        try {
+            val pendingSwap = breezSDK.inProgressSwap()
+            Logger.tag(TAG).info { "Found pendingSwap: ${pendingSwap}" }
+            notifyChannel(
+                context,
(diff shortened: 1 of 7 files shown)


ok300:
> Explicitly try to redeem the swap funds (complete swap)
> #778

The most direct way to do that is to call `breez_sdk_core::swap_in::swap::redeem_swap(bitcoin_address)`. However that's a private method.

The way I found to get to it is to call `breezSDK.inProgressSwap()`. This internally calls `execute_pending_swaps()` which eventuall calls `redeem_swap()`.

@roeierez what do you think?

roeierez:
> > Explicitly try to redeem the swap funds (complete swap)
> > #778
> 
> The most direct way to do that is to call `breez_sdk_core::swap_in::swap::redeem_swap(bitcoin_address)`. However that's a private method.
> 
> The way I found to get to it is to call `breezSDK.inProgressSwap()`. This internally calls `execute_pending_swaps()` which eventuall calls `redeem_swap()`.
> 
> @roeierez what do you think?

I think we should expose explicitly `redeem_swap` as notifications may be given on old swaps (not in progress).

android/app/src/main/kotlin/com/cBreez/client/job/RedeemSwap.kt
```diff
@@ -0,0 +1,55 @@
...
+    private val context: Context,
+    private val fgService: ForegroundService,
+    private val payload: String,
+) : SDKJob {
+    companion object {
+        private const val TAG = "RedeemSwapJob"
+    }
+
+    override fun start(breezSDK: BlockingBreezServices) {
+        val request = Json.decodeFromString(AddressTxsConfirmedRequest.serializer(), payload)
```
erdemyerebasmaz, line 32: Is thing going to be used with `redeemSwap` API?

erdemyerebasmaz (commented):
Changes look structurally good!

Have one comment about an unused variable, which I presume will be used with `redeemSwap` once it's exposed through SDK.

erdemyerebasmaz:
Rebased branch on top of `main` branch to fix CI issues.
- Flutter 3.19 changes
- Breaking SDK change: Use `ConnectRequest` param on `connect` API 
- _Optional SDK branch selection on CI/CD workflow_

ok300:
The PR is ready for final review.

dangeross (approved):
LGTM

android/app/src/main/kotlin/com/cBreez/client/job/RedeemSwap.kt
```diff
@@ -0,0 +1,54 @@
...
+
+        try {
+            breezSDK.rescanSwaps()
```
roeierez, line 33: This is a very heavy call as it goes over all historical addresses and I think we don't want to use it within the job that is resources and time limited.
We need to find a way to refresh this specific address information.

roeierez (changes requested):

android/app/src/main/kotlin/com/cBreez/client/job/RedeemSwap.kt
ok300: Would this be a good solution? breez/breez-sdk/pull/823

ok300: Call to `rescanSwaps` removed: e05f412412efe6cde7df5c3d6411861489dddf3c

roeierez (approved):

PR #785, opened 2024-02-16
Fix the serialization of the LNURL-pay responses

dangeross:
The response field names should be camel cased


roeierez (approved):
LGTM

erdemyerebasmaz (approved):

PR #790, opened 2024-02-20
Environment variables workaround for Flutter 3.19

erdemyerebasmaz:
`--dart-define-from-file` had the ability to get access to provided key-value pairs in gradle and plist but it was _silently_ removed with **Flutter 3.19**. It was not considered a breaking change and was not documented in release notes by Flutter team.

I was not able to catch this on my builds, perhaps due to cache issues. @dangeross encountered this issue and found the related issues on this information & a [workaround](https://github.com/flutter/flutter/issues/142976#issuecomment-1949602247).

### Changelist:
- Apply the workaround so that the dart defined variables are available from `gradle` 448d34a29bd99498e649e55c44791c26173bd372
  - The iOS counterpart isn't implemented as we're not using environment variables on our Notification Service App Extension, they are accessed through shared Keychain
- In light of this, removed previously added environment variables from iOS & removed the related step from Build iOS workflow 13eb1a388768f94102749f868435bdb5408e5178
- Update `troubleshooting.md` to include `--dart-define` examples as well faf9599839183de92f76b5c0761ccffbf028d98f

### Other Changelist:
- Depend on the main **Firebase** modules instead of the KTX modules 2d63a9007c479c564144f5bd2b6c1df39b3b8000
  - More information here: [Migrate to using the Kotlin extensions (KTX) APIs in the main modules](https://firebase.google.com/docs/android/kotlin-migration)
- Removed unused **WorkManager** `gradle` dependency fcefb4d8f6bf56605185dc8bc75b7010765c536d

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -130,11 +130,10 @@ flutter {
 
 dependencies {
     /* FCM */
-    implementation 'com.google.firebase:firebase-messaging-ktx:23.4.0'
-    /* WorkManager */
     def work_version = "2.9.0"
     // Kotlin + coroutines
     implementation "androidx.work:work-runtime-ktx:$work_version"
+    implementation 'com.google.firebase:firebase-messaging:23.4.1'
     /* Logging */
     implementation 'org.tinylog:tinylog-api-kotlin:2.6.2'
     implementation 'org.tinylog:tinylog-impl:2.6.2'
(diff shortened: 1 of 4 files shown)


sbatezat:
See also https://github.com/flutter/flutter/issues/139289#issuecomment-1954103555

ademar111190:
Terrible thing from the flutter team; great we catch up this

roeierez:
@erdemyerebasmaz We still need the api key passed to the app. how is the iOS API Key is propagated to the app now?

samz1943:
any work around for this --dart-define-from-file?

PR #796, opened 2024-02-21
Use recommended fees when calculating claim tx feerates for reverse swap fee estimates

### LINKED ISSUE #791: Send to BTC Address: Correctly calculate fees based on chosen feerate
Currently, the "Send to BTC Address" is incorrectly using `prepare_redeem_onchain_funds` to calculate the reverse swap fees.

It should instead use `fetch_reverse_swap_fees` and https://github.com/breez/breez-sdk/issues/801 to calculate the actual reverse swap fees.

### LINKED ISSUE #793: Send to BTC Address: Different behavior of "Use all funds" flag vs Breez Mobile
On Breez Mobile, on the "Send to BTC Address" screen:
- ticking the "Use all funds" field allows me to advance to the next screen
- not ticking "Use all funds", but manually inputting the maximum amount of funds available results in an error "Breez requires you to keep 1 sat in your balance"

On c-breez, on the same screen, both ticking and not ticking "Use all funds" has the same behavior.

erdemyerebasmaz:
example workflow

This PR
fixes #791
fixes #793

### Changelist:
- Fee calculations are done separately on each swap bloc(`ReverseSwapBloc`, `RedeemOnchainFundsBloc`, `RefundBloc`) 93fcdedb86df377dfaf2bdadb24d49dd389c0d47

---

##### FeeOption changes:
- `FeeOptionsBloc` & `FeeOptionsState` are removed
  - `FeeOption` is moved under models directory
- `FeeOption` is now declared as an abstract class and each swap has their own `FeeOption`:
  - `ReverseSwapFeeOption` , `RedeemOnchainFeeOption`, `RefundFeeOption`
  - Reverse Swap, Redeem Onchain & Refund flows now utilize these FeeOptions
- `waitingTime` is now part of `ProcessingSpeed` enum
- Check if account has enough balance/onchain balance when checking if fees are affordable

---

##### Reverse Swap changes:
- Use recommended fees when calculating claim tx feerates for reverse swap fee estimates.
- Calculate the boltz fees in a more accurate manner
  - so that the input amount is the amount received on target address
- Pass the (swap amount + calculated fees) to sendOnchain request

**Notes:** Check whether updating reverse swap fee options before opening reverse swap confirmation page to make sure we have a valid feesHash is necessary.

---

- Refactor `FeeChooser` widgets to be compatible with all FeeOptions 5ad43577bc71134d90f2e1688d74fdad69b75251
  - Rename `walletBalance` to `amountSat`
  - Use walletBalance for `RedeemOnchainFeeOption`
  - Handle "Use All Funds" cases properly. 
  - `FeeBreakdown` is broken down into widgets for readability
  - Calculate boltz service fees internally on `FeeBreakdown`
    - Make boltzServiceFee calculation more readable 4bcd62dadd2accf8642cb8b1e5e2d517a9d03527
- Convert `ProcessingSpeedWaitTime` into a Widget 39b47e5f6d837385498941a89408d90ddb223634
- Give anyhow errors precedence over other errors when extracting their error message 35ae74aa867ad554fa62e73caa41bf4a98999f39
- Update dependencies to latest

---

### Regression tests on:
- [ ] Redeem Onchain_(Unexpected Funds)_ flow
- [ ] Refund flow


lib/models/fee_options/fee_option.dart
```diff
@@ -0,0 +1,91 @@
...
+    required this.pairInfo,
+    required super.txFeeSat,
+    required super.processingSpeed,
+    required super.satPerVbyte,
+  });
+
+  @override
+  bool isAffordable({required int balance, required int amountSat, bool? isMaxValue}) {
+    if (isMaxValue == true) {
+      return (amountSat + boltzServiceFee(amountSat) + pairInfo.feesClaim) > 0;
```
roeierez, line 51: Isn't this always true?

lib/routes/withdraw/reverse_swap/confirmation_page/reverse_swap_confirmation_page.dart
```diff
@@ -90,7 +87,10 @@ class _ReverseSwapConfirmationPageState extends State<ReverseSwapConfirmationPag
 
   void _fetchReverseSwapFeeOptions() {
     final reverseSwapBloc = context.read<ReverseSwapBloc>();
-    _fetchFeeOptionsFuture = reverseSwapBloc.fetchReverseSwapFeeOptions(sendAmountSat: widget.amountSat);
+    _fetchFeeOptionsFuture = reverseSwapBloc.fetchReverseSwapFeeOptions(
+      amountSat: widget.amountSat,
+      amountType: widget.isMaxValue ? SwapAmountType.Receive : SwapAmountType.Send,
```
ok300, line 92: Shouldn't this be the other way around?

```dart
amountType: widget.isMaxValue ? SwapAmountType.Send : SwapAmountType.Receive
```

erdemyerebasmaz: 🤔 I don't think so.
If we're sending maximum value, we'd like to drain our balance or send maximum amount within policy. 
Let's say our balance is 58k, we're interested in getting fees for senderAmountSat:58k.

lib/routes/withdraw/reverse_swap/confirmation_page/reverse_swap_confirmation_page.dart
```diff
@@ -57,44 +52,61 @@ class _ReverseSwapConfirmationPageState extends State<ReverseSwapConfirmationPag
...
                 )
               : null,
     );
   }
+
+  void _fetchReverseSwapFeeOptions() {
+    final reverseSwapBloc = context.read<ReverseSwapBloc>();
+    _fetchFeeOptionsFuture = reverseSwapBloc.fetchReverseSwapFeeOptions(
+      amountSat: widget.amountSat,
+      amountType: widget.isMaxValue ? SwapAmountType.Receive : SwapAmountType.Send,
```
roeierez, line 92: Shouldn't this be the other way around? I mean other than if the user wants to drain all the funds they want to input the recipient amount. Isn't it the way in breezmobile?

roeierez (changes requested):

lib/routes/withdraw/reverse_swap/confirmation_page/reverse_swap_confirmation_page.dart
roeierez: If the user wants to drain all channel (e.g 58k) then the user wants to say "I want to send 58k how much do I receive at the end after fees)"
So the for max amount the type should be `Send`.

erdemyerebasmaz: You're both correct. I made my comment looking at the snippet Carlos attached as the correct example, not the implementation 😅 

Fixed c9cf3750b1b76f6065c3b1535992c6ee20b903d9

lib/models/fee_options/fee_option.dart
erdemyerebasmaz: I guess I wanted to write balance - (fees) > 0 but it's the same with the statement below.

I wasn't able to give this much attention while working on notifications. I'm reviewing & testing this whole thing again with fresh eyes.

erdemyerebasmaz:
Applied changes from:
- https://github.com/breez/breez-sdk/pull/850

example workflow

lib/routes/withdraw/widgets/fee_chooser/widgets/fee_chooser_header.dart
```diff
@@ -35,13 +38,19 @@ class _FeeChooserHeaderState extends State<FeeChooserHeader> {
...
-              isSelected: widget.selectedFeeIndex == index,
-              onSelect: () => widget.onSelect(index),
-            ),
+            (index) {
+              var feeOption = widget.feeOptions.elementAt(index);
+              return FeeOptionButton(
+                index: index,
+                text: feeOption.getDisplayName(texts),
+                isAffordable: feeOption.isAffordable(
+                  balance: (feeOption is RedeemOnchainFeeOption) ? account.walletBalance : account.balance,
```
erdemyerebasmaz, line 47: ⚠️ 
Redeem & Refund should use wallet balance.

erdemyerebasmaz: Done 433729d599a5cc31442bc8ccd02f1dea1772c1e9

ok300 (approved):

erdemyerebasmaz:
[RevSwap v2 changes](https://github.com/breez/c-breez/compare/ffae419e18763daea3cfcecbc3efe14e92a60e8d...reverse_swap_optional_fees)

PR #798, opened 2024-02-22
iOS Swap Tx Notification

dangeross:
Adds iOS handling of address_txs_confirmed notification message


android/app/src/main/kotlin/com/cBreez/client/job/LnurlPayInfo.kt
```diff
@@ -58,16 +59,16 @@ class LnurlPayInfoJob(
...
+                fail(e.message, request.replyURL)
+            }
             notifyChannel(
                 context,
                 Constants.NOTIFICATION_CHANNEL_LNURL_PAY,
                 context.getString(R.string.lnurl_pay_notification_failure_title),
             )
         }
-
-        fgService.shutdown()
```
erdemyerebasmaz, line 71: Why was this removed? Is it because `LnurlPayInvoice` task is expected to start soon afterwards?

What if we don't have enough inbound liquidity to pay the lnurl invoice? I don't think we do receive a _"lnurlpay_invoice"_ notification for rejected/failed cases.

dangeross: My bad, sorry. Reverted the change

erdemyerebasmaz (approved):
LGTM

PR #802, opened 2024-02-27
LNURL-Payment: Channel setup fee limit

### LINKED ISSUE #777: Automatic channel limit
With Lnurlpay we currently only return maxSendable as the inbound liquidity which means such received payments are limited to that.
We want to expand the ability to pay via lnurl pay (and other methods that require automatic invoice creation) also in cases when a channel is required.
In order to do so we need to add a setting for automatic channel creation fee limit and in the lnurl response we should do the following:
1. Change maxSendable to the maximum payment we can get (including openning channels if needed).
2. In the invoice request we try to create an invoice and then:

- If the opening channel fee return is more than the user settings we respond with appropriate error
- If the invoice can not be created due to not enough fees to cover the channel we respond with error.

erdemyerebasmaz:
Closes #777

This PR allows users to set a "`Automatic channel setup fee limit"` through **Lightning Fees** page to be able to receive LNURL payments that require a channel setup where the fees fall within that limit. 
On C-Breez, this feature is **enabled** & set to **5000 sats** by default.

Breez SDK PR:
- https://github.com/breez/breez-sdk/pull/833

### Todo:
- If the invoice can not be created due to not enough fees to cover the channel respond with error.

@dangeross @roeierez
Currently, on the extensions we're getting the value from shared preferences but this approach is not directly compatible with extension plugin. To be able to make this compatible with extension plugin, I think a good approach would be:
- Make this part of SDK `Config`
  - Should be disabled and set to 0 by default imo, I'm not sure if this is a decision we should make on behalf of the SDK users
- Use the _automatic channel setup fee limit_ from `Config` that was used on `connect` when setting `maxSendable`
  - This value needs to be accessible internally

### DIFF
--- a/android/app/src/main/kotlin/com/cBreez/client/BreezForegroundService.kt
+++ b/android/app/src/main/kotlin/com/cBreez/client/BreezForegroundService.kt
@@ -1,5 +1,6 @@
 package com.cBreez.client
 
+import android.content.SharedPreferences
 import breez_sdk.ConnectRequest
 import breez_sdk.EnvironmentType
 import breez_sdk.GreenlightNodeConfig
@@ -9,6 +10,7 @@ import breez_sdk.defaultConfig
 import breez_sdk.mnemonicToSeed
 import breez_sdk_notification.ForegroundService
 import breez_sdk_notification.NotificationHelper.Companion.registerNotificationChannels
+import breez_sdk_notification.ServiceConfig
 import com.breez.breez_sdk.SdkLogInitializer
 import com.cBreez.client.utils.FlutterSecuredStorageHelper.Companion.readSecuredValue
 import io.flutter.util.PathUtils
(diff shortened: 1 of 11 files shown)


roeierez:
How about defining a config object for the plugin and pass it in the intent?
This way every app cal define this config whenever it wants and feed it to the plugin.

erdemyerebasmaz:
Extension logic has been moved to: 
- https://github.com/breez/c-breez/pull/804
- https://github.com/breez/c-breez/pull/806

erdemyerebasmaz:
> @erdemyerebasmaz can you please rebase this? Will be easier to review

Rebased

erdemyerebasmaz:
[UI changes](https://github.com/breez/c-breez/compare/main...f03492f092743083fd275585485fa8b9532b2440)

Extension changes on 
- Android: 68baa180b4f6ddc7afbc6520d40d70ab2cdbf495
- iOS: 8bc0d11c0158852f38b84407954c3fe0eba4f5e5

Issue #771, opened 2024-01-25
Use lnurl pay domain on payment details dialog for payments not made to a LN Address


erdemyerebasmaz:
Related SDK PR:
- https://github.com/breez/breez-sdk/pull/735

PR #815, opened 2024-03-12
Use lnurl pay domain on payment details dialog for payments not made to a LN Address

erdemyerebasmaz:
Fixes https://github.com/breez/c-breez/issues/771


roeierez (approved):
LGTM

PR #817, opened 2024-03-13
Reduce default log level from `CONFIG` to `INFO`

ok300:
The `CONFIG` level resulted in logging the sending and receiving of gRPC data frames, which is spamming the logs:

```
[BreezSdk] {CONFIG} (2024-03-13T09:51:00.634293Z) : send frame=Data { stream_id: StreamId(1) }
[BreezSdk] {CONFIG} (2024-03-13T09:51:00.735160Z) : send frame=Data { stream_id: StreamId(1), flags: (0x1: END_STREAM) }
[BreezSdk] {CONFIG} (2024-03-13T09:51:01.103566Z) : received frame=Settings { flags: (0x1: ACK) }
[BreezSdk] {CONFIG} (2024-03-13T09:51:01.117228Z) : received settings ACK; applying Settings { flags: (0x0), enable_push: 0, initial_window_size: 2097152, max_frame_size: 16384 }
[BreezSdk] {CONFIG} (2024-03-13T09:51:01.157191Z) : received frame=WindowUpdate { stream_id: StreamId(0), size_increment: 73 }
[BreezSdk] {CONFIG} (2024-03-13T09:51:01.222232Z) : received frame=Headers { stream_id: StreamId(1), flags: (0x4: END_HEADERS) }
[BreezSdk] {CONFIG} (2024-03-13T09:51:01.271599Z) : received frame=Data { stream_id: StreamId(1) }
[BreezSdk] {CONFIG} (2024-03-13T09:51:01.346240Z) : received frame=Headers { stream_id: StreamId(1), flags: (0x5: END_HEADERS  | END_STREAM) }
```

These logs appear to be emitted by the lower-level networking calls in flutter. I found no way to specifically target this networking module with a higher log level. @erdemyerebasmaz @ademar111190 if you see a better solution, please let me know.

PR #825, opened 2024-03-14
Get the correct setup fee limit from SharedPreferences

erdemyerebasmaz:
- Adds "flutter." prefix to SharedPreferences key for setup fee limit on extensions
  - This is the [default prefix](https://github.com/flutter/packages/blob/b21c5424783a9e74cb841a8df7a15f26eba04a7b/packages/shared_preferences/shared_preferences/lib/shared_preferences.dart#L18) used by shared_preferences package


android/app/src/main/kotlin/com/cBreez/client/BreezForegroundService.kt
```diff
@@ -39,8 +39,8 @@ class BreezForegroundService : ForegroundService() {
                 "ERROR" -> Logger.tag(TAG).error { l.line }
                 "WARN" -> Logger.tag(TAG).warn { l.line }
                 "INFO" -> Logger.tag(TAG).info { l.line }
-                "DEBUG" -> Logger.tag(TAG).debug { l.line }
-                "TRACE" -> Logger.tag(TAG).trace { l.line }
+                // "DEBUG" -> Logger.tag(TAG).debug { l.line }
+                // "TRACE" -> Logger.tag(TAG).trace { l.line }
```
dangeross, line 43: Do these need reseting?

erdemyerebasmaz: No, it's continuation of
- #817 
but I'll address it as part of another PR.

erdemyerebasmaz:
There's [an issue on iOS](https://stackoverflow.com/a/76355960) where we can't access shared_preferences values from the extension. Addressing this by using [shared preferences with App Group](https://pub.dev/packages/shared_preference_app_group) plugin when setting channel setup fee limit.

lib/utils/preferences.dart
```diff
@@ -65,6 +66,8 @@ class Preferences {
     _log.info("set payment options channel setup limit fee : $channelFeeLimitMsat");
     final prefs = await SharedPreferences.getInstance();
     await prefs.setInt(_kPaymentOptionChannelSetupFeeLimit, channelFeeLimitMsat);
+    // iOS Extension requirement
+    await SharedPreferenceAppGroup.setInt(_kPaymentOptionChannelSetupFeeLimit, channelFeeLimitMsat);
```
dangeross, line 70: Is there a need to set it twice? We could also get the preference in the same way using `SharedPreferenceAppGroup.get(_kPaymentOptionChannelSetupFeeLimit)`

dangeross: Makes sense now sorry, it's iOS only

dangeross (approved):
LGTM

erdemyerebasmaz:
- Refactored PaymentOptionsPage & PaymentOptionsBloc as there were a lot of unwanted behaviors due to bad state management practices.
- Removed PaymentOptionsBloc tests temporarily.

lib/utils/preferences.dart
erdemyerebasmaz: Yes, this is workaround for not being able to access `shared_preferences` values on iOS extension. Values set through `shared_preference_app_group` live on app group's UserDefaults and do not overlap with `shared_preferences` values.

PR #828, opened 2024-03-15
Set initial default shared preferences

dangeross:
This PR changes the `PaymentOptionsBloc` to set the initial default shared preferences (in iOS also to the app group UserDefaults) if they are currently not stored. This allows the notification service extension to get the default values from shared preferences.

### DIFF
--- a/ios/Runner.xcodeproj/project.pbxproj
+++ b/ios/Runner.xcodeproj/project.pbxproj
@@ -10,11 +10,11 @@
 		1498D2341E8E89220040F4C2 /* GeneratedPluginRegistrant.m in Sources */ = {isa = PBXBuildFile; fileRef = 1498D2331E8E89220040F4C2 /* GeneratedPluginRegistrant.m */; };
 		2000AD59359B91B3C7963B30 /* LnurlPayInvoice.swift in Resources */ = {isa = PBXBuildFile; fileRef = 590A05D747C4C91AE3779479 /* LnurlPayInvoice.swift */; settings = {ASSET_TAGS = (BreezSDK, ); }; };
 		2B214792005A6D185AF0B44C /* Pods_Runner.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = FACDE338D59177DD56404E3D /* Pods_Runner.framework */; };
-		35F2EA3A2C522E5BCF3B705A /* ServiceConfig.swift in Resources */ = {isa = PBXBuildFile; fileRef = D5089940E3A84AD08707A5AE /* ServiceConfig.swift */; settings = {ASSET_TAGS = (BreezSDK, ); }; };
 		3B3967161E833CAA004F5970 /* AppFrameworkInfo.plist in Resources */ = {isa = PBXBuildFile; fileRef = 3B3967151E833CAA004F5970 /* AppFrameworkInfo.plist */; };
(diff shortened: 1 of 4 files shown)


lib/bloc/payment_options/payment_options_bloc.dart
```diff
@@ -29,22 +29,26 @@ class PaymentOptionsBloc extends Cubit<PaymentOptionsState> {
...
+      } catch (e) {
+        _log.severe("Failed to reset payment options.");
+        rethrow;
+      }
     }
+
+    await getPaymentOptions();
   }
 
   Future<void> getPaymentOptions() async {
```
erdemyerebasmaz, line 54: `resetPaymentOptions` is being called from a UI widget but we can make `getPaymentOptions` a private method as it's not used elsewhere.

lib/bloc/payment_options/payment_options_bloc.dart
```diff
@@ -11,7 +11,7 @@ class PaymentOptionsBloc extends Cubit<PaymentOptionsState> {
   PaymentOptionsBloc(
     this._preferences,
   ) : super(const PaymentOptionsState.initial()) {
-    getPaymentOptions();
+    resetPaymentOptions(true);
   }
```
erdemyerebasmaz, line 15: Realized we can simplify this whole logic to:

```suggestion
        _initializePaymentOptions();
  }

  void _initializePaymentOptions() async {
    await _preferences.hasPaymentOptions() ? await getPaymentOptions() : await resetPaymentOptions();
  }
```

ios/Runner.xcodeproj/project.pbxproj
erdemyerebasmaz, line 1: Why were these changes included? Are they necessary?

erdemyerebasmaz (changes requested):
Have one suggestion... We made it more complex than it is :)

lib/bloc/payment_options/payment_options_bloc.dart
```diff
@@ -29,22 +29,26 @@ class PaymentOptionsBloc extends Cubit<PaymentOptionsState> {
...
-      _log.info(
-        "Reverted payment options to default values: "
-        "proportionalFee: $kDefaultProportionalFee "
-        "exemptFeeMsat: $kDefaultExemptFeeMsat "
-        "channelFeeLimitMsat: $kDefaultChannelSetupFeeLimitMsat",
-      );
-    } catch (e) {
-      _log.severe("Failed to reset payment options.");
-      rethrow;
+  Future<void> resetPaymentOptions([bool checkPaymentOptions = false]) async {
```
erdemyerebasmaz, line 32: `resetPaymentOptions` can be renamed to `setDefaultPaymentOptions` as we're not really resetting anything when shared preferences does not have payment options.

ios/Runner.xcodeproj/project.pbxproj
dangeross: It wasn't building without adding `ServiceConfig` to the compile sources

lib/bloc/payment_options/payment_options_bloc.dart
dangeross: Looks a lot simpler now

erdemyerebasmaz (approved):
LGTM

PR #830, opened 2024-03-18
Correct fee affordability calculations for swaps

erdemyerebasmaz:
Corrects the fee affordability calculations for refund & redeem onchain.
- Checks if refund amount is greater or equal to tx fees,
- Checks if onchain balance is greater or equal to redeem onchain amount.

I'm not satisfied with the shared Fee widgets & models on swap/reverse swap and plan to go over them to make them more robust at a later date.

### DIFF
--- a/lib/models/fee_options/fee_option.dart
+++ b/lib/models/fee_options/fee_option.dart
@@ -32,9 +32,7 @@ abstract class FeeOption {
     }
   }
 
-  bool isAffordable({required int balance, required int amountSat}) {
-    return (amountSat + txFeeSat) < balance;
-  }
+  bool isAffordable({int? balance, int? walletBalance, required int amountSat});
 }
 
 class ReverseSwapFeeOption extends FeeOption {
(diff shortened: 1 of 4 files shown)


lib/models/fee_options/fee_option.dart
```diff
@@ -32,9 +32,7 @@ abstract class FeeOption {
     }
   }
 
-  bool isAffordable({required int balance, required int amountSat}) {
-    return (amountSat + txFeeSat) < balance;
-  }
+  bool isAffordable({int? balance, int walletBallance, required int amountSat});
```
roeierez, line 35: NIT: 
balance => lightningBalance 
walletBalance =< onchainBalance

lib/models/fee_options/fee_option.dart
```diff
@@ -67,4 +70,9 @@ class RedeemOnchainFeeOption extends FeeOption {
     required super.processingSpeed,
     required super.satPerVbyte,
   });
+
+  @override
+  bool isAffordable({int? balance, int? walletBallance, required int amountSat}) {
```
roeierez, line 75: If `walletBallance` is required then we should check if not null and return false.

roeierez (changes requested):

lib/models/fee_options/fee_option.dart
erdemyerebasmaz: These are the naming we use on app:
https://github.com/breez/c-breez/blob/0895c06664119fc9bdb8095a1f6da6611648fa15/lib/bloc/account/account_state_assembler.dart#L26-L27

lib/models/fee_options/fee_option.dart
erdemyerebasmaz: Done.

roeierez (approved):

PR #857, opened 2024-03-27
Set "https://mempool.space/" as default mempool url on app layer

erdemyerebasmaz:
Applies minimum required changes for Breez SDK PR:
- https://github.com/breez/breez-sdk/pull/898

These changes do integrate the fallback behavior but we still rely on "https://mempool.space/" to display the transactions onchain on the UI.

That is until the user sets a Mempool API URL of their choice from **Preferences** -> **Network**.

### DIFF
--- a/lib/config.dart
+++ b/lib/config.dart
@@ -15,12 +15,12 @@ class Config {
 
   final sdk.Config sdkConfig;
   final sdk.NodeConfig nodeConfig;
-  final String defaultMempoolUrl;
+  final String? defaultMempoolUrl;
 
   Config._({
     required this.sdkConfig,
     required this.nodeConfig,
-    required this.defaultMempoolUrl,
+    this.defaultMempoolUrl,
   });
 
   static Future<Config> instance({
@@ -33,7 +33,7 @@ class Config {
       final breezSDK = injector.breezSDK;
       final breezConfig = await _getBundledConfig();
       final defaultConf = await _getDefaultConf(breezSDK, breezConfig.apiKey, breezConfig.nodeConfig);
-      final defaultMempoolUrl = defaultConf.mempoolspaceUrl;
+      final defaultMempoolUrl = defaultConf.mempoolspaceUrl ?? "https://mempool.space/";
       final sdkConfig = await getSDKConfig(injector, defaultConf, breezConfig);
 
       _instance = Config._(
@@ -99,7 +99,7 @@ class Config {
     return configuredExemptFee;
   }
 
-  static Future<String> _mempoolSpaceUrl(
+  static Future<String?> _mempoolSpaceUrl(
     ServiceInjector serviceInjector,
     sdk.Config defaultConf,
   ) async {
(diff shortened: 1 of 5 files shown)


ok300:
The SDK will automatically fetch, cache then use a list of mempool.space instances, if none is provided from the outside.

So I'd say this is not necessary.

erdemyerebasmaz:
> The SDK will automatically fetch, cache then use a list of mempool.space instances, if none is provided from the outside.
> 
> 
> 
> So I'd say this is not necessary.

SDK isn't provided with mempoolspace_url unless the user specifically sets one from Network preferences page.

We only use "mempool.space" by default on app level to have valid blockchain url's such as:
image

lib/config.dart
```diff
@@ -33,7 +33,7 @@ class Config {
       final breezSDK = injector.breezSDK;
       final breezConfig = await _getBundledConfig();
       final defaultConf = await _getDefaultConf(breezSDK, breezConfig.apiKey, breezConfig.nodeConfig);
-      final defaultMempoolUrl = defaultConf.mempoolspaceUrl;
+      final defaultMempoolUrl = defaultConf.mempoolspaceUrl ?? "https://mempool.space/";
```
ok300, line 36: If the app explicitly sets it, only that instance will be used in the SDK, which will bypass the automatic fallback introduced in https://github.com/breez/breez-sdk/pull/898 .

I'd suggest to only set it if the user sets it.

ok300: Also: the new base URL format is `https://mempool.space/api`.

Is that something the app can handle? (e.g. append `/api` and handle the trailing slashes?)

erdemyerebasmaz: > If the app explicitly sets it, only that instance will be used in the SDK, which will bypass the automatic fallback introduced in [breez/breez-sdk#898](https://github.com/breez/breez-sdk/pull/898) .
> 
> I'd suggest to only set it if the user sets it.

This `defaultMempoolUrl` is only used by application config and not passed sdk config.

> Also: the new base URL format is `https://mempool.space/api`.
> 
> Is that something the app can handle? (e.g. append `/api` and handle the trailing slashes?)

It's handled here: [BlockChainExplorerUtils](https://github.com/breez/c-breez/blob/main/lib/utils/blockchain_explorer_utils.dart) and tested.

ok300: So if the user sets the mempool.space URL in the app settings, it won't be given as arg when initializing the SDK?

erdemyerebasmaz: > So if the user sets the mempool.space URL in the app settings, it won't be given as arg when initializing the SDK?

It will be passed to SDK only if explicitly set from app settings, if not SDK will treat it as empty and use the fallback mechanism.

ok300: Asking because, with the change in https://github.com/breez/breez-sdk/pull/898 , it's now possible for the user to set their own mempool.space URL if so desired.

ok300: Sounds good, thanks.

ok300 (approved):

lib/config.dart
ok300: > It will be passed to SDK only if explicitly set from app settings, if not SDK will treat it as empty and use the fallback mechanism.

Just as reminder, when and if it's set in the SDK, it has to be in the `https://mempool.space/api` format.

PR #858, opened 2024-03-27
Refactor swap in progress

erdemyerebasmaz:
#### Changelist:
- Revise swap in progress bloc & state
  - Ensure the bloc isn't closed when emitting states
  - Don't stop polling even if swap in progress is waiting for confirmation
-  SwapPage: Swap in progress widget takes precedence over deposit address widget

### DIFF
--- a/lib/bloc/swap_in_progress/swap_in_progress_bloc.dart
+++ b/lib/bloc/swap_in_progress/swap_in_progress_bloc.dart
@@ -34,7 +34,8 @@ class SwapInProgressBloc extends Cubit<SwapInProgressState> {
       _log.info("Refreshing swap in address.");
       _emitState(state.copyWith(isLoading: true));
       SwapInfo? inProgress = (await _breezSDK.inProgressSwap());
-      _emitState(state.copyWith(inProgress: inProgress));
+      // Reset unused if there's a swap in progress
+      _emitState(state.copyWith(inProgress: inProgress, unused: inProgress != null ? null : state.unused));
       if (state.inProgress == null && state.unused == null) {
         _emitState(
           state.copyWith(
--- a/lib/bloc/swap_in_progress/swap_in_progress_bloc.dart
+++ b/lib/bloc/swap_in_progress/swap_in_progress_bloc.dart
@@ -13,63 +13,64 @@ final _log = Logger("SwapInProgressBloc");
 class SwapInProgressBloc extends Cubit<SwapInProgressState> {
   final BreezSDK _breezSDK;
 
-  SwapInProgressBloc(this._breezSDK) : super(SwapInProgressState(null, null, isLoading: true)) {
-    pollSwapAddress();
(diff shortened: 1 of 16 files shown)


lib/bloc/swap_in_progress/swap_in_progress_bloc.dart
```diff
@@ -13,63 +13,64 @@ final _log = Logger("SwapInProgressBloc");
...
-        swapUnused =
-            currentState.unused ?? (await _breezSDK.receiveOnchain(req: const ReceiveOnchainRequest()));
-      }
-      _log.info("swapInProgress: $swapInProgress, swapUnused: $swapUnused");
-      if (!isClosed) {
-        emit(SwapInProgressState(swapInProgress, swapUnused));
+      _log.info("Refreshing swap in address.");
+      _emitState(state.copyWith(isLoading: true));
+      SwapInfo? inProgress = (await _breezSDK.inProgressSwap());
+      _emitState(state.copyWith(inProgress: inProgress));
```
roeierez, line 37: If inProgress != null then we should clean the unused from the state shouldn't we?
Otherwise when the swap completed we will still have there the same unused value instead of calling receive_onchain again.

roeierez (changes requested):

lib/bloc/swap_in_progress/swap_in_progress_bloc.dart
erdemyerebasmaz: Yes, missed it when making state changes.
Done c1eaf6545a29dcb605299b6375263400a8aedd5a

lib/bloc/swap_in_progress/swap_in_progress_bloc.dart
```diff
@@ -13,63 +13,64 @@ final _log = Logger("SwapInProgressBloc");
...
-        swapUnused =
-            currentState.unused ?? (await _breezSDK.receiveOnchain(req: const ReceiveOnchainRequest()));
-      }
-      _log.info("swapInProgress: $swapInProgress, swapUnused: $swapUnused");
-      if (!isClosed) {
-        emit(SwapInProgressState(swapInProgress, swapUnused));
+      _log.info("Refreshing swap in address.");
+      _emitState(state.copyWith(isLoading: true));
+      SwapInfo? inProgress = (await _breezSDK.inProgressSwap());
+      _emitState(state.copyWith(inProgress: inProgress, unused: null));
```
roeierez, line 37: I think we only need to set unused to null if inProgress is not null. Currently it seems that if there is no in progress swap the receivce_onchain is invoked on every iteration.

roeierez (changes requested):

lib/bloc/swap_in_progress/swap_in_progress_bloc.dart
erdemyerebasmaz: Corrected.
Reset unused to null if there's a swap in progress 4d4ef41

PR #871, opened 2024-03-29
Fix background polling on swap pages

erdemyerebasmaz:
- Regression: Replace function reference with function call
- Remove async logic
- Hide loader for background refreshes


roeierez (approved):
LGTM

PR #872, opened 2024-03-29
Remove helper methods to print SDK classes

erdemyerebasmaz:
These helpers methods are no longer needed as we're able to print SDK structs with the changes introduced in Breez SDK PR:
- https://github.com/breez/breez-sdk/pull/910


roeierez (approved):
LGTM

PR #885, opened 2024-04-14
Address LateInitializationError on swap timers

### LINKED ISSUE #875: Fix edge case where late initialization error can happen on swap timers
If first call to `_refreshInProgressReverseSwaps` or `_refreshAddresses` hangs without an error due to network issues timer would have not been initialized and this would cause a LateInitializationError as we're leaving these pages and blocs try to stop polling(_that has never got started_).

erdemyerebasmaz:
Closes #875


dangeross (approved):
LGTM

PR #900, opened 2024-07-16
Allow payments to be made if there's a callback domain mismatch with SuccessAction URL

erdemyerebasmaz:
This PR applied changes from
- https://github.com/breez/breez-sdk/pull/1039

so that callback domain validation is bypassed for LNURL-Pay success action URL's and callback domain and C-Breez behavior is the same with Breezmobile.


roeierez (approved):
LGTM

Issue #888, opened 2024-04-26
Application crashes in iOS 15.x


dangeross:
I've seen this crash now recently when using the application on iPhone 7 iOS 15.8.2 and also tested further on the simulator. The application functions up until you restore from mnemonic (SDK connects) and then immediately crashes from then on.

From looking at the crash report it seems there is a dynamic link library missing, most likely libsqlite3 (on calling `rusqlite::statement::Statement::execute_with_bound_parameters`)

Devices: iPhone 7, iPhone 12 (sim), iPhone 13 (sim)
OS: iOS 15.0, iOS 15.2

```
Thread 34 Crashed:: frb_workerpool
0   dyld                          	       0x117208c66 __abort_with_payload + 10
1   dyld                          	       0x117222fd7 abort_with_payload_wrapper_internal + 82
2   dyld                          	       0x117223009 abort_with_payload + 9
3   dyld_sim                      	       0x10f875060 abort_with_payload + 26
4   dyld_sim                      	       0x10f83508b dyld4::halt(char const*) + 375
5   dyld_sim                      	       0x10f84d032 dyld4::APIs::_dyld_missing_symbol_abort() + 34
6   Runner                        	       0x10d3df749 rusqlite::statement::Statement::execute_with_bound_parameters::h7bf0f0f5ef5a8940 + 87
7   Runner                        	       0x10d3dac79 rusqlite::Connection::execute::h743bd44db761d106 + 151
8   Runner                        	       0x10d45427b breez_sdk_core::persist::db::SqliteStorage::get_connection::h8ba9632908dd7210 + 235
9   Runner                        	       0x10d453dd3 breez_sdk_core::persist::db::SqliteStorage::init::he79733dfdab25dfa + 700
10  Runner                        	       0x10d413d86 breez_sdk_core::binding::connect::_$u7b$$u7b$closure$u7d$$u7d$::h4e46bee66c517c2a + 1271
11  Runner                        	       0x10d36d2bc tokio::runtime::park::CachedParkThread::block_on::hf3c0f1a7449488bf + 153
12  Runner                        	       0x10d386a1d tokio::runtime::context::blocking::BlockingRegionGuard::block_on::hca9c9a41c0feb7c4 + 18
13  Runner                        	       0x10d37d9a2 tokio::runtime::context::runtime::enter_runtime::h3f86f5c1f996aed6 + 157
[...]

PR #908, opened 2024-09-30
iOS 15.x crash

erdemyerebasmaz:
This PR addresses crashes specific to iOS 15.x due to conflict between system-provided & our own sqlite3 implementation.


dangeross:
Fixes #888

dangeross (approved):

PR #918, opened 2025-04-23
fix the swap-in refund sender amount

JssDWt:
The tx fee for a swap-in should be deducted from the amount. The sender amount is `amountSat` here.


roeierez (approved):
LGTM

